PR #46, opened 2018-07-20
Remove MAT dependency (fixes #42)

### LINKED ISSUE #42: Remove MAT-package dependency

maxbennedich:
- Created a small Serialization module to write and read sparse matrices to text files on disk.
- Replaced qdep_infbilanczos.mat with text files.
- Replaced gun.mat with text files.
- Created gun_native test to test the gun data. Input is welcome -- does this test make sense by itself, and what should we do with the old gun test?
- Moved fetch_nlevp_path to GalleryNLEVP, since it was no longer used in Gallery.
- Tried compressing the text files via ZipFile, since they're quite large to have checked in (9.5 MB), but the decompression was painfully slow. Let me know if you think this is a problem.


jarlebring:
Nice. We anticipate more converted mat-files from the NLEVP collection, so I suggest we put them in a separate directory e.g. `gallery_extra/converted_nlevp/gun_K`,  etc.  10 MB is okay with me. The gun_native test looks good. Since we have reference solutions for the gun-problem, we could also add a test which checks that `qnewton` actually computes a reference eigenvalue up to some small error.  I think there are reference solutions stored in the `gun.mat`, or you can get them with say nleigs.

maxbennedich:
Good suggestions. I updated the code. I didn't find any reference eigenvalues in `gun.mat`, but used nleigs. Do you think we should add all known reference eigenvalues, and make sure that the one that the test finds is one of them, or just keep the eigenvalue that the test ends up finding? I did the latter for now.

eringh:
I think it looks good.

If there are many files we want to store like this it may be worth putting these in a separate repo to avoid the package being too large. But that might be a later choice.

Another thing one could do is to put some kind of extra test to compare with the MATLAB implementation of NLEVP, but exclude it from the regular tests. On the other hand, maybe those will never be run again and if we find the same eigenvalue I guess it is good enough.

I think we can merge this!

maxbennedich:
Ok, thanks! Yes, putting large resources in a separate repository could be a good idea, to reduce the size of our package.

Issue #51, opened 2018-08-15
Sum of NEPs


jarlebring:
Several NEPs in applications are naturally formed as sums of other NEPs, potentially of different NEP-types. (Quite close to @maxbennedich's recent work in NEPTypes for NLEIGS, who could use a PEP and SPMF)

It can be achieved by
```
struct SUMNEP{NEP1<:NEP,NEP2<:NEP}  <: NEP 
   nep1::NEP1
   nep2::NEP2
end
```
which allows us to to delegate interface functions, e.g.,
``` 
  compute_Mlincomb(nep::SUMNEP, λ, V) = (compute_Mlincomb(nep1, λ, V)+compute_Mlincomb(nep2, λ, V))
```
For the sum of two NEPs, this seems natural and seems to be doable in a type-stable way. (Note that the types of nep1 and nep2 is stored in the parameter of SUMNEP.) However, the fact that this is only two NEPs is hard-coded. Sometimes one would like to construct the sum of several NEPs. I don't know how to do that in a type stable way. 

Attempt 1: The struct
```
struct SUMNEP  <: NEP 
   nepv::Vector{<:NEP}
end
```
allows storing a sum of more than two NEPs, but does not contain the type information in the parameter of `SUMNEP` (so not good for type stability). 
Attempt 2: The definition 
```
struct SUMNEP{COMMONNEP<:NEP}  <: NEP 
   nepv::Vector{COMMONNEP}
end
```
would store type information, but it would force the NEPs to be of the same type. 

Any other ideas how to handle the sum of more than two NEPs in a type stable way?

eringh:
One trivial and dirty way is to simply implement it as the first suggestion, i.e.,
```
struct SUMNEP{NEP1<:NEP,NEP2<:NEP}  <: NEP 
   nep1::NEP1
   nep2::NEP2
end
```
and then just let the delegation work its magic. Since a `SUMNEP` may contain another `SUMNEP`. In such case one could also think of overloading the `+`-operator to create a `SUMNEP` as the sum of two `NEPs`. It is an easy way out, but I doubt it will be efficient.

eringh:
In the case of two `NEPs` of the same type being summed it could be interesting to create a new `NEP` of the same type. It would also be interesting to preserve abstract types such as `ProjectableNEP` and `AbstractSPMF`.

But that is for later. Just adding to a wish list of what a `SUMNEP` should do, rather than how.

PR #52, opened 2018-08-15
SUMNEP type #51

jarlebring:
Illustration how  `SUMNEP` type can be added.  @maxbennedich, do you think this could be integrated with your `PNEP` type?  This example shows at least how one can create a sum of SPMF and PEP:
```
  julia> nep1=PEP([randn(2,2),randn(2,2),randn(2,2)])
  julia> A0=[1 3; 4 5]; A1=[3 4; 5 6];
  julia> id_op=S -> eye(S)
  julia> exp_op=S -> expm(S)
  julia> nep2=SPMF_NEP([A0,A1],[id_op,exp_op]);
  julia> sumnep=SumNEP(nep1,nep2);
```

### DIFF
--- a/src/Gallery.jl
+++ b/src/Gallery.jl
@@ -387,8 +387,8 @@ module Gallery
           nep = nep_gallery("nlevp_native_gun")
           K, M = get_Av(nep.nep1);
           W1, W2 = get_Av(nep.nep2);
-          c1 = LowRankMatrixAndFunction(W1, nep.fi[3])
-          c2 = LowRankMatrixAndFunction(W2, nep.fi[4])
+          c1 = LowRankMatrixAndFunction(W1, get_fv(nep.nep2)[1])
+          c2 = LowRankMatrixAndFunction(W2, get_fv(nep.nep2)[2])
           return SumNEP(PEP([K, M]), LowRankFactorizedNEP([c1, c2]))
       else
           error("The name $name is not supported in NEP-Gallery.")
--- a/src/Gallery.jl
+++ b/src/Gallery.jl
@@ -376,19 +376,22 @@ module Gallery
           M=read_sparse_matrix(gunbase * "M.txt")
           W1=read_sparse_matrix(gunbase * "W1.txt")
           W2=read_sparse_matrix(gunbase * "W2.txt")
-          minusop= S-> -S
+          minusop= S-> S
           oneop= S -> eye(size(S,1),size(S,2))
           sqrt1op= S -> 1im*sqrtm(Matrix(S))
           sqrt2op= S -> 1im*sqrtm(Matrix(S)-108.8774^2*eye(S))
-          AA=[K,M,W1,W2];
-          nep=SPMF_NEP(AA,[oneop,minusop,sqrt1op,sqrt2op])
+          AA=[K,-M,W1,W2];
+          pep=PEP([K,-M]);
+          sqrtnep=SPMF_NEP([W1,W2],[sqrt1op,sqrt2op]);
+          nep=SumNEP(pep,sqrtnep);
           return nep;
       elseif name == "nlevp_gun_low_rank_nep"
           nep = nep_gallery("nlevp_native_gun")
-          K, M, W1, W2 = nep.A
+          K, M = get_Av(nep.nep1);
+          W1, W2 = get_Av(nep.nep2);
           c1 = LowRankMatrixAndFunction(W1, nep.fi[3])
           c2 = LowRankMatrixAndFunction(W2, nep.fi[4])
-          return SumNEP(PEP([K, -M]), LowRankFactorizedNEP([c1, c2]))
+          return SumNEP(PEP([K, M]), LowRankFactorizedNEP([c1, c2]))
       else
           error("The name $name is not supported in NEP-Gallery.")
       end
--- a/src/Gallery.jl
+++ b/src/Gallery.jl
@@ -383,13 +383,6 @@ module Gallery
           sqrtnep=SPMF_NEP([W1,W2],[sqrt1op,sqrt2op]);
           nep=SumNEP(pep,sqrtnep);
           return nep;
-      elseif name == "nlevp_gun_low_rank_nep"
-          nep = nep_gallery("nlevp_native_gun")
-          K, M = get_Av(nep.nep1);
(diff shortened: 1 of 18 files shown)


maxbennedich:
Hmm, yes that might work, but it wouldn't remove the need for a new type by itself right, since for NLEIGS we also need the nonlinear matrix LU factors and ranks? (The new type I added converts the problem to a single SPMF, which it wraps together with attached LU factors and some metadata.) We may also miss out on optimizations that can be done by having a single type, for example the sparsity pattern pre-allocation that can be done for SPMFs by combining all matrices, although I'm not sure how big of a difference that makes in practice.

jarlebring:
You are right. We would need one additional NEP-type.  The following could define the non-polynomial part for what is now the `PNEP``:
```
type LUFactorizedNEP <: AbstractSPMF
   Av::Vector{<:AbstractMatrix}
   fv::Vector{Function}
   Lv::Vector{<:AbstractMatrix}  
   Uv::Vector{<:AbstractMatrix}
end
```
In this construction, what is now `PNEP` would correspond to `nep::SumNEP{PEP,LUFactorizedNEP}`. I think the metadata can be extracted?  e.g., polynomial degree:  `size(nep.nep1.A,1)`. The rank is just `size(Lv[i],1)`? 

I currently feel slightly in favor of such a solution since the `LUFactorizedNEP` could go NEP-algorithm code eg `method_nleigs.jl` rather than `NEPTypes.jl`. I think it would make the interface code in `NEPTypes.jl` cleaner and be less algorith-specific. I can see how `SumNEP` is useful for other NEP-algorithms, whereas a `LUFactorizedNEP` and `PNEP` less so. Other NEP-methods have specialized NEP-types (e.g. in Broyden's method).

maxbennedich:
Ah, yes that makes sense to keep `NEPTypes` clean and keep the algorithm-specific code out of it. An alternative is to make `PNEP` specific to NLEIGS and rename/move it there. I experimented with a few different types before I ended up with `PNEP`, including something similar to `LUFactorizedNEP`, and I couldn't figure out a good way to handle the compute methods, since I wanted to reuse what is already implemented for `SPMF_NEP`. Perhaps you can think of a better solution, but what I ended up doing thus was to wrap a `SPMF_NEP` and simply forward the compute methods there.

maxbennedich:
I've updated the NLEIGS Gun problem to use SumNEP. Please take a look and let me know what you think @jarlebring . If we proceed with this approach, the remaining work as I see it is:

- Move `LowRankFactorizedNEP` and the `AbstractMatrixAndFunction` types from `NEPTypes.jl` to the NLEIGS directory
- Support other types of SumNEPs, notably `SumNEP{PEP, SPMF}`
- Update remaining NLEIGS tests to use SumNEP

jarlebring:
It looks good. I updated the `SumNEP`-type a bit to "handle" the fact that the sum of two SPMFs should be seen as a AbstractSPMF. This way I could change `nlevp_native_gun` to return a `SumNEP` (or actually a `SPMFSumNEP`) without breaking tests. I think `SumNEP{PEP,SPMF}` now works as expected at least for test/gun_native and test/blocknewton. 

Since we for now want to move `LowRankFactorizedNEP` to method_nleigs we need to remove it from Gallery. I think we can now move the code in  gallery `nlevp_gun_low_rank_nep` to method_nleigs, or the unit test for nleigs+gun.

Do you think it is difficult to do the following? Separate the two modes of nleigs separated as two method-definitions 
```
function nleigs(
        ::Type{T},
        nep::NEP,
        Sigma::AbstractVector{CT};
        kwargs...)
```
and
```
function nleigs(
        ::Type{T},
        nep::SumNEP{PEP,LowRankFactorizedNEP},
        Sigma::AbstractVector{CT};
        kwargs...)
```
It would allow for julia dispatch, and make the code quite clear that there are two modes.

maxbennedich:
That's seems like a good solution. Note that there are at least three modes of NLEIGS which could make sense to support, ordered from least to most efficient:

1. Any NEP that supports M(λ) (`NEP`)
2. PEP + SPMF, without low rank exploitation (`SPMFSumNEP{PEP,SPMF}`)
3. PEP + SPMF, with low rank exploitation (`SPMFSumNEP{PEP,LowRankFactorizedNEP}`)

We could also support a fourth version which is any `SPMF` (which is the same as version 2 with an empty PEP part). This can be handled more efficiently than version 1, since the NLEIGS algorithm can still use the matrices and functions, but if there is a PEP part, then version 2 will be better. The only concern is that we want to make sure users use the most efficient/specific version possible, and don't accidentally use e.g. the SPMF version when their problem could be stated as a PEP + SPMF.

maxbennedich:
Hmm, I'll take a look. I'll sync it with master too to resolve the merge conflict.

maxbennedich:
That specific problem is resolved, however there is now a new situation in that `compute_Mder` (and perhaps other compute methods) fail when one NEP is empty. We could support empty NEPs, alternatively not allow SumNEPs with empty NEPs (you could argue that it shouldn't be a SumNEP then). Opinions?

jarlebring:
I managed to make empty SPMFs work together with SumNEP: 
```
julia> nep1=SPMF_NEP(3);
julia> nep2=SPMF_NEP(3);
julia> nep=SumNEP(nep1,nep2);
julia> compute_Mder(nep,1.0)
3×3 SparseMatrixCSC{Float64,Int64} with 0 stored entries
```
Since `LowRankFactorizedNEP` is delegating to the nep.spmf. I would think it is doable by an appropriate initiation of `LowRankFactorizedNEP`. Maybe the commented line in this constructor could help? (I will not be able to work more on this today.)
```
    function LowRankFactorizedNEP(n) # Create an empty LowRankFactorizedNEP
        #LowRankFactorizedNEP(SPMF_NEP(n),0,Vector{Matrix{Float64}}(0),Vector{Matrix{Float64}}(0))
        nep=LowRankFactorizedNEP(Vector{MatrixAndFunction{Matrix{Float64}}}(0)) #
        return nep
    end
```

maxbennedich:
Ah, yes that helped, thanks!

maxbennedich:
I fixed the remaining tests, and moved out the low rank code to a new module in the nleigs folder. I'll merge this PR now, and we can revisit later if something else comes up.

PR #53, opened 2018-08-15
Cleanup compute_Mlincomb

jarlebring:
In this PR I try to clean up the code for `compute_Mlincomb`.   Some comments below (which can later be used as a basis in manuscript). These are valid calls to `compute_Mlincomb`:
```
compute_Mlincomb(nep,λ,V)
compute_Mlincomb(nep,λ,V,a)
compute_Mlincomb(nep,λ,V,a,startder)
compute_Mlincomb!(nep,λ,V)
compute_Mlincomb!(nep,λ,V,a)
compute_Mlincomb!(nep,λ,V,a,startder)
```
where `λ::Number, V::Union{AbstractMatrix,AbstractVector}, a::Vector, startder::Integer`.

I did not like the `@method_concretely_defined` calls. The code to automatically select a procedure for compute_Mlincomb using `@method_concretely_defined` (by calls to e.g. `compute_Mlincomb_from_MM`) was not very good for performance and other reasons. I managed to remove it, at a cost: as a NEP-type developer you have to make explicit call to e.g., `compute_Mlincomb_from_MM`.  For instance, for PEPs this required the addition of one definition
```
    compute_Mlincomb(nep::PEP,λ::Number,
                     V::Union{AbstractMatrix,AbstractVector},a::Vector=ones(size(V,2)))=
             compute_Mlincomb_from_MM(nep,λ,V,a)
```

I also decided to remove kwargs for the `compute_Mlincomb` since it lead to a code repetition. It had a kwarg for `a::Vector` which is a scaling vector. This is now an optional parameter. Similar for the `startder::Integer`. 

Instructions:

**As a _algorithm-developer_ using `compute_Mlincomb`:** 
If you want to form a linear combination of derivative matrices, you should try to specify the scaling vector `a` rather than scaling the columns yourself. 
(a) If, after your call, your algorithms does `V` any more, call  `compute_Mlincomb!`
(b) If you need the contents of `V` to be unchanged after the call, use `compute_Mlincomb`.

Situation (a) is preferred.

**As a _type-developer_ (beginner user):**

Implement `size` and `compute_Mder`. Make the `compute_MM` and `compute_Mlincomb` available by defining:
```    
compute_Mlincomb(nep::MyNEP,λ::Number,V::Union{AbstractMatrix,AbstractArray})=compute_Mlincomb_from_Mder(nep,λ,V)
compute_MM(nep::MyNEP,S::Matrix,V::Matrix)=compute_MM_from_Mder(nep,S,V)
```
Accept it may not be extremely efficient. 

**As a _type-developer_ (novice user):**

Implement `size` and `compute_MM`. Make the `compute_Mder` and `compute_Mlincomb` available by defining:
```    
compute_Mlincomb(nep::MyType,λ::Number,V::Union{AbstractMatrix,AbstractVector},a::Vector=ones(size(V,2)))=compute_Mlincomb_from_MM(nep,λ,V,a)
compute_Mder(nep::MyType,...)=compute_Mder_from_MM()
```

**As _type-developer_ (Advanced user):** 
* implement `size`,  `compute_MM`,  `compute_Mder` and
  a) `compute_Mlincomb`, if you are not modifying the contents of `V`; or
  b) `compute_Mlincomb!`, if you are modifying `V`.
 Situation (a) is preferred. 
* If bottleneck in your approach, implement  also `compute_Mlincomb(nep::MyType,λ::Number,V::Union{AbstractMatrix,AbstractVector},a::Vector)` and `compute_Mlincomb(nep::MyType,λ::Number,V::Union{AbstractMatrix,AbstractVector},a::Vector,startder:Integer)`.

### DIFF
--- a/src/Gallery.jl
+++ b/src/Gallery.jl
@@ -376,19 +376,13 @@ module Gallery
           M=read_sparse_matrix(gunbase * "M.txt")
           W1=read_sparse_matrix(gunbase * "W1.txt")
           W2=read_sparse_matrix(gunbase * "W2.txt")
-          minusop= S-> -S
-          oneop= S -> eye(size(S,1),size(S,2))
+          # The gun problem is a sum of a PEP and a problem containing square roots.
+          pep=PEP([K,-M]);
           sqrt1op= S -> 1im*sqrtm(Matrix(S))
           sqrt2op= S -> 1im*sqrtm(Matrix(S)-108.8774^2*eye(S))
-          AA=[K,M,W1,W2];
-          nep=SPMF_NEP(AA,[oneop,minusop,sqrt1op,sqrt2op])
+          sqrtnep=SPMF_NEP([W1,W2],[sqrt1op,sqrt2op]);
+          nep=SumNEP(pep,sqrtnep);
           return nep;
-      elseif name == "nlevp_gun_pnep"
-          nep = nep_gallery("nlevp_native_gun")
-          K, M, W1, W2 = nep.A
-          c1 = LowRankMatrixAndFunction(W1, nep.fi[3])
-          c2 = LowRankMatrixAndFunction(W2, nep.fi[4])
-          return PNEP([K, -M], [c1, c2])
       else
           error("The name $name is not supported in NEP-Gallery.")
       end
--- a/src/LinSolvers.jl
+++ b/src/LinSolvers.jl
@@ -100,7 +100,7 @@ module LinSolvers
 
         function GMRESLinSolver{T_num, T_nep}(nep::T_nep, λ::T_num, kwargs) where {T_num<:Number, T_nep<:NEP}
             function f(v::AbstractVector)
-              return compute_Mlincomb(nep, λ, v, a=[1])
+              return compute_Mlincomb(nep, λ, v)
             end
             A = LinearMap{T_num}(f, size(nep,1), ismutating=false)
             gmres_log = false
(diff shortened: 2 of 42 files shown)


jarlebring:
Suggestion about exclamation mark in NEPCore:
```
compute_Mlincomb!(nep::NEP,params...)=compute_Mlincomb(nep,params...)
```

jarlebring:
In commit d6ec86cf75aaad993b04b3c9e4185e12a843d3c5 and fe20b19796bc96fe201394dfaf9b4999394878c9 I added code for default behavior for `compute_Mlincomb!` and `compute_Mlincomb`. Note that 
* `compute_Mlincomb(::NEP,V,..)` points to `compute_Mlincomb!(..,copy(V),..)` 
* `compute_Mlincomb!(::NEP,V.. )` points to `compute_Mlincomb(..V..)`.  

i.e., they point to each other. If a type-developer implements one NEP-type specific method (either with or without exclamation mark) both will work.   (I found it quite entertaining that the wrappers can point to each other in this way.)

As a type-developer, `compute_Mlincomb` is preferred and as a NEP-algorithm developer `compute_Mlincomb!` is preferred. I updated the PR-description to say this.

eringh:
Nice (and interesting)!

I guess it works since a soon as there are more specialized functions available Julia will choose these. I suspect that if you define a new type and do not implement any of the Mlincomb functions you will get a stack overflow if you try to call them. Compare with the example
```
f(a) = g(a)
g(a) = f(a)
function f(a::Int64)
  return a + 1
end
function g(a::Float64)
  return a - 1.0
end
println(f(1))
println(g(1))
println(f(1.0))
println(g(1.0))
f(1.0+1.0im)
```
Maybe highlight (and stress) the requirements for a beginner user since otherwise the error produced will potentially look strange.

PR #54, opened 2018-08-21
Faster SPMF computation

maxbennedich:
There seems to be a bug in the current implementation of the sparsity pattern `Zero` matrix for SPMF NEPs. This PR is a suggestion for a fix, and is the fastest solution I've come up with for summing sparse matrices. It works by first determining the sparsity pattern of all matrices combined, and then storing each matrix using that combined sparsity pattern. This makes the multiply/addition step very efficient since it can be done without any memory allocation at all, by just summing the `nzval` vectors.

For the Gun problem (4 matrices), this makes `compute_Mder` 4x faster than the current implementation, and uses 6x less memory. For the Particle problem, there's a huge improvement, since the problem consists of 81 nonlinear matrices, all with identical sparsity pattern, making `compute_Mder` ~70x faster. (This was the main motivation for this PR, since `compute_Mder` is a performance bottleneck for that problem.)

For matrices with very different sparsity patterns, there is a risk that this change has a negative impact, since all matrices would be stored with the combined sparsity pattern and thus use more space and operations. One could imagine making this method optional (a flag when creating the NEP), or calculating the overhead by dividing the combined `nnz` with the average matrix `nnz` and don't use this method if the factor is larger than, say, 3.

This PR is not quite ready to be merged yet, as I wanted to get some feedback first if this change makes sense, and if there are any problems, or improvements that can be made. If we decide to go ahead with this PR, what remains to be done is:

- Not hard code the type Complex128
- Optional sparsity pattern?
- Support sparsity pattern in other compute methods?
- Use this method for all matrices of `SPMFSumNEP`s (e.g. PEP + SPMF)?

### DIFF
--- a/src/NEPTypes.jl
+++ b/src/NEPTypes.jl
@@ -259,7 +259,7 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
 
     function compute_Mder(nep::SPMF_NEP,λ::Number,i::Integer=0)
         if i == 0
-            x = map(i -> nep.fi[i](reshape([λ],1,1))[1], 1:length(nep.As))
+            x = map(i -> nep.fi[i](reshape([λ],1,1))[1], 1:length(nep.fi))
 
             if isempty(nep.As)
                 Z = copy(nep.Zero)
--- a/src/NEPTypes.jl
+++ b/src/NEPTypes.jl
@@ -266,9 +266,7 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
                 n = length(Z.nzval)
                 @inbounds for a = 2:length(nep.As)
                     x = T(nep.fi[a](reshape([λ],1,1))[1])
-                    for i = 1:n
-                        Z.nzval[i] += nep.As[a].nzval[i] * x
-                    end
+                    Z.nzval .+= nep.As[a].nzval .* x
                 end
                 return Z
             end
--- a/src/NEPTypes.jl
+++ b/src/NEPTypes.jl
@@ -119,11 +119,11 @@ for matrices in the standard matrix function sense.
 
          # Sparse zero matrix to be used for sparse matrix creation
          Zero::SparseMatrixCSC
-         As::Vector{SparseMatrixCSC{Complex128,Int}}  # A matrices with sparsity pattern of all matrices combined
+         As::Vector{SparseMatrixCSC{<:Number,Int}}  # 'A' matrices with sparsity pattern of all matrices combined
     end
 
     SPMF_NEP(n, A, fi, Schur_factorize_before, Zero) =
-        SPMF_NEP(n, A, fi, Schur_factorize_before, Zero, Vector{SparseMatrixCSC{Complex128,Int}}())
+        SPMF_NEP(n, A, fi, Schur_factorize_before, Zero, Vector{SparseMatrixCSC{Float64,Int}}())
 
 """
      SPMF_NEP(AA, fii, Schur_fact = false, use_sparsity_pattern = true)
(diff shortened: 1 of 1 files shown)


src/NEPTypes.jl
```diff
@@ -165,18 +170,37 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
...
+
+                 for col = 1:size(A, 2)
+                    for j in nzrange(A, col)
+                        S[A.rowval[j], col] = A.nzval[j]
+                    end
+                 end
+
+                 push!(As, S)
              end
-             Zero=(Zero*1im)*0
```
maxbennedich: This was the bug. This will drop all zeros, resulting in an empty matrix.

jarlebring:
Nice work. I will try to do a review (never done it in git).

src/NEPTypes.jl
```diff
@@ -228,11 +252,23 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
...
+                end
+                return Z
+            else
+                T = Complex{Float64}
+                x = T(nep.fi[1](reshape([λ],1,1))[1])
+                Z = copy(nep.As[1])
+                Z.nzval .*= x
+                @inbounds for a = 2:length(nep.As)
+                    x = T(nep.fi[a](reshape([λ],1,1))[1])
+                    Z.nzval .+= nep.As[a].nzval .* x
```
jarlebring: The way to handle types here is a bit tricky. I would say `T` should be determined from the "greatest" type of `eltype(nep.Av[i])` and `typeof(nep.fi[i](reshape([λ],1,1))`, e.g., if one of them is `Complex{Float64}` but all others are `Float64` then `T=Complex{Float64}`.  I think "greatest" can be determined with [the `promote_rule` commands](https://docs.julialang.org/en/v0.6.2/manual/conversion-and-promotion/). We would maybe first need to compute a vector `Vector{Any}` of all the fi-values.

jarlebring:
> Use this method for all matrices of SPMFSumNEPs (e.g. PEP + SPMF)?

I would say this technique can be propagated to `PEP`s in a natural way.  How to adapt this for `SumNEP` would also maybe be nice but less straightforward. I don't see how to this without making SPMF parametric on the matrices.

maxbennedich:
> I don't see how to this without making SPMF parametric on the matrices.

Yes, regarding this, is this desirable? I.e. enforcing parametrically that all matrices are the same type (both in terms of the matrix type and the element type)? From an implementation standpoint I think this would be nice, since it feels like it'd lead to safer and simpler code. But would there be cases where mixing types (say sparse + full or real + complex matrices) is preferred?

jarlebring:
If a user wants to mix dense and sparse matrices (or maybe more commonly `Diagonal` or `Tridiagonal`) matrices, it can still be achieved now that we have `SumNEP`. One puts dense terms in `nep1::SPMF_NEP{Matrix{Float64}}` and sparse matrices in `nep2::SPMF_NEP{SparseMatrixCSC{Float64}}` and letting `nep=SumNEP(nep1,nep2)`.

It would be okay with me to do `SPMF_NEP` parametric, now that we understand it better. It seems it should be done carefully (similar to `DEP` and check type stability) so maybe thats a different PR.

src/NEPTypes.jl
```diff
@@ -228,11 +252,24 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
...
-                Z=Z+nep.A[i]*nep.fi[i](reshape([λ],1,1))[1]
+            if isempty(nep.As)
+                Z=copy(nep.Zero)
+                for i=1:size(nep.A,1)
+                    Z=Z+nep.A[i]*nep.fi[i](reshape([λ],1,1))[1]
+                end
+                return Z
+            else
+                T = Complex{Float64}
+                x = T(nep.fi[1](reshape([λ],1,1))[1])
```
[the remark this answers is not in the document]
maxbennedich: It should be taken care of by the `if isempty(nep.As)` check a few lines above.

src/NEPTypes.jl
maxbennedich: Should be fixed now, without loss of performance it seems.

maxbennedich:
I've addressed the comments, and added an optional `use_sparsity_pattern` flag. Anything remaining before merging?

jarlebring:
Looks good.

PR #69, opened 2018-09-07
Jd deflation

eringh:
I start to feel ready with this implementation. The work contains an implementation of the Effenberger deflated version of the Jacob-Davidson algorithm, based on robust expansion of invariant pairs.

There are two TODOs marked, although I think it is not worth prioritizing. The first, regarding the error measure I am not sure how to do. Compare with the equation (9)/(3.2) in the paper by Effenberger. I am unsure if we can use any other error measure.
The second regards how to continue the method from an already computed basis after convergence of an eigenvalue. This requires some assumptions on the inner solver, or some code around it, to extract the information needed.

I also contains some minor changes in other files related to type-stability and bugs.


eringh:
The pull request fails the build since master is broken.

maxbennedich:
I'll fix master in a few minutes.

eringh:
I was trying to do that, but I had problems running Julia 1.0 and problems with the package when running Julia 0.7.

maxbennedich:
Alright, I pushed an attempted fix (restoring the previous code), once it builds successfully you could try re-syncing with master.

src/NEPTypes.jl
```diff
@@ -853,13 +853,13 @@ julia> compute_Mder(nep,λ)[1:2,1:2]
...
-        else
-            return n
-        end
+    function size(nep::Proj_NEP,dim)
+        n = size(nep.W,2);
+        return n
+    end
+    function size(nep::Proj_NEP)
+        n = size(nep.W,2);
+        return (n,n)
```
maxbennedich: Nice!

eringh:
I just synced with master to see that it works. Since I'm having trouble with Julia 1.0 locally.

eringh:
Since everything seems to work I'll merge with master.

PR #75, opened 2018-09-13
Legacy removal 1

eringh:
I have purged the repo of some old and unused code.
It is the MATLABLinSolves which fell out when we removed the MATLAB dependency. We do not need them, and they were not used anywhere in the project.
Moreover it is the "bugs", "tmp", "extra_tests", and "trash" folders. These have not been touched for quite a while and is out of date.

In an earlier commit (c4294dbd2a1487cd8889a9e462992cf3237af5ed) I stared to remove old and unused tests from "extra_tests".

This PR is mostly for documentation.


maxbennedich:
Great! The best code is no code at all. We can always refer back to the repository history if we need it in the future.

PR #77, opened 2018-09-13
Fiber native working (with newton interpolation)

jarlebring:
I have managed to make nlevp fiber work natively:
```
  nep=nep_gallery("nlevp_native_fiber")
```
It was much trickier than I thought because of singularities of Bessel K functions at zero. Also, we don't have bessel matrix functions, so I needed to handle it some different way. This solution is interpolation and creating a rational function. The rational function can be evaluated in a matrix sense, so it now returns an spmf (where the last term is a combination of matrix square root and  a rational term evaluated in with a Newton-polynomial scheme).  

The not so nice aspect is that the approach is very dependent on interpolation points. There is a singularity at λ=0 and the closest eigenvalue is 7.13949325474625e-7. I found an interpolation which works for that, but you may argue that I cheated since I did some tuning of the interval used for interpolation. 

(I will update PR a bit with comments and more tests.)

### DIFF
--- a/src/Gallery.jl
+++ b/src/Gallery.jl
@@ -27,53 +27,53 @@ module Gallery
 \\
 
 \\
-  **Supported 'name' and 'params':**\\
-     'dep0'\\
-      Create a random delay eiganvalue problem with one delay tau = 1\\
+  **Supported "name" and 'params':**\\
+\\
+     `dep0`\\
+Create a random delay eiganvalue problem with one delay tau = 1\\
       * one optional parameter determining the size (default = 5)\\
 \\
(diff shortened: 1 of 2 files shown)


eringh:
Great!

Singularities around zero is something that is tricky in may problems. The Gun problem has a singularity in the derivatives at zero.
```
nep = nep_gallery("nlevp_native_gun");

julia> any(isnan.(compute_Mlincomb(nep,0.0,ones(size(nep,1)),[1],1)))
true

julia> any(isnan.(compute_Mlincomb(nep,0.0,ones(size(nep,1)),[1],0)))
false
```

Issue #59, opened 2018-09-01
Extend projection matrices


jarlebring:
When we construct projected problems we currently create a `Proj_NEP` and the projecion is subsequently modified by `set_projection_matrices!(pnep,X,Y)`. In most NEP-methods where we need a projected solve,  we extend the projection matrices one column at a time. In this setting the computation of the projection matrices can be done more efficiently. 

I suggest we add an "interface function"
```
   extend_project_matrices!(pnep,x::AbstractVector,y::AbstractVector)
```
which is mathematically the same as doing
```
   set_project_matrices!(pnep,[X,x],[Y,y])
```
where X and Y are the current project matrices.

Issue #70, opened 2018-09-07
Rewrite of compute_MM(SPMF 


jarlebring:
As pointed out in #68, `compute_MM(::SPMF` is not efficient. The type of "optimization" we did for `compute_MM(::DEP` also needs to be done with a general SPMF. Similar to the DEP we need to make SPMF parametric with respect to `AbstractMatrix`. Here, I think we also need to make it parametric with respect to the internal type of the functions provided. Otherwise we can never have type stability, since the compiler would not be able to determine the output type of `fi(S)`. 

So, the SPMF would be something like this:
```
    struct SPMF_NEP{Ftype<:Number,  T<:AbstractMatrix} <: AbstractSPMF
         n::Int
         A::Vector{T}   # Array of Array of matrices
         fi::Vector{Function}  # Array of functions
.... 
```
The type logic of the function is that the functions `f(S)` should always return a matrix with greatest type of `Ftype` and `eltype(S)`. 

**updated**
Rewrite includes:
* [x] Make compute_Mlincomb type stable 
* [x] Independent implementantion on compute_Mder  (not relying on compute_Mder_from_MM)
* [x] Make compute_Mder type stable 
* [x] Make compute_MM type stable
* [x] Make SPMF parametric
* [ ] Reporting that it gets better

eringh:
Sounds like a good idea if we want type stability. Question is if the JIT compiler would be able to see the return statements and compile for it.
Maybe something like [this](https://docs.julialang.org/en/latest/manual/performance-tips/#Annotate-values-taken-from-untyped-locations-1) could be used by the method developers.
Or maybe take a look into something like [this](https://github.com/yuyichao/FunctionWrappers.jl).

maxbennedich:
Agreed, sounds like a good idea! Getting everything type stable sounds a bit tricky if the calculation needs to promote type depending on input, but maybe achievable. Do we need to support different types within the vectors of matrices or functions? For example that one matrix or function could be complex, and another not?

`FunctionWrappers` looks very interesting, it's exactly what I've been missing in Julia! Would be interesting to see if we can apply it in any of our code to get better type stability.

jarlebring:
FunctionWrappers would have to be carefully looked in particular wrt maturity.

>  Getting everything type stable sounds a bit tricky if the calculation needs to promote type depending on input, but maybe achievable

[...]

PR #80, opened 2018-09-14
Proj_NEP update

jarlebring:
This is a preparation for #59 where we want to eventually extend projected problems. This PR changes the storage of the `Proj_SPMF_NEP`. For memory reasons the projected NEP is created from the `pnep.projnep_B_mem` which is an vector of (large) matrices. The projected problems are created as SubArrays of this, when `set_proj_matrices!` is called.

This required changes in `nlar`, `tiar`, `jd`. 

Based on the benchmark report from travis, this PR creates a slight regression in nlar and jd, but I think we will take this back once we actually implement an `extend_proj_matrices!`: 
Before PR:
```
  Nonlinear Arnoldi                                          59.2 s    12.7 GiB
  Jacobi–Davidson                                            1.99 s     460 MiB
```
After PR:
```
  Nonlinear Arnoldi                                          64.5 s    13.2 GiB
  Jacobi–Davidson                                            2.02 s     467 MiB
```

### DIFF
--- a/src/NEPTypes.jl
+++ b/src/NEPTypes.jl
@@ -241,7 +241,18 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
                 end
             end
             ## Sum it up
-            Z=Z+nep.A[i]*(V*Fi);
+
+
+            VFi=V*Fi;
+            if (isa(nep.A[i],SubArray) && (eltype(nep.A[i]) != eltype(VFi)))
+                # SubArray x Matrix of different types do not work?
+                # https://discourse.julialang.org/t/subarray-x-matrix-multipliciation-of-different-eltype-fails/14950
+                # Workaround by making a matrix copy
+                Z=Z+copy(nep.A[i])*VFi;
+            else
+                Z=Z+nep.A[i]*VFi;
+            end
+
         end
         return Z
     end
@@ -754,6 +765,7 @@ where ``M(λ)`` is represented by `orgnep`. Use
         nep_proj::SPMF_NEP; # An instance of the projected NEP
         orgnep_Av::Vector
         orgnep_fv::Vector
+        projnep_B_mem::Vector # A vector of matrices
         function Proj_SPMF_NEP(nep::AbstractSPMF,maxsize::Int,T)
             this = new(nep)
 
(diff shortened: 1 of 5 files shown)


src/NEPTypes.jl
```diff
@@ -241,7 +241,18 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
                 end
             end
             ## Sum it up
-            Z=Z+nep.A[i]*(V*Fi);
+
+
+            VFi=V*Fi;
+            if (isa(nep.A[i],SubArray) && (eltype(nep.A[i]) != eltype(VFi)))
```
eringh: Maybe put a "TODO" here. Just to remember that we should come back and check it every now and then.

src/NEPTypes.jl
```diff
@@ -804,12 +828,16 @@ julia> compute_Mder(nep,3.0)[1:2,1:2]
         ## Sets the left and right projected basis and computes
         ## the underlying projected NEP
         m = size(nep.orgnep_Av,1);
-        B = Array{Array{eltype(W),2}}(undef, m);
+        T=eltype(eltype(nep.projnep_B_mem));
+
+        T_sub = SubArray{T,2,Array{T,2},Tuple{UnitRange{Int64},UnitRange{Int64}},false}
```
eringh: Maybe `Matrix{T}` instead of `Array{T,2}`?

eringh: Or perhaps: Maybe create the first B[i] and determine the type based on that.

src/NEPTypes.jl
```diff
@@ -241,7 +241,18 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
...
             ## Sum it up
-            Z=Z+nep.A[i]*(V*Fi);
+
+
+            VFi=V*Fi;
+            if (isa(nep.A[i],SubArray) && (eltype(nep.A[i]) != eltype(VFi)))
+                # SubArray x Matrix of different types do not work?
+                # https://discourse.julialang.org/t/subarray-x-matrix-multipliciation-of-different-eltype-fails/14950
+                # Workaround by making a matrix copy
+                Z=Z+copy(nep.A[i])*VFi;
```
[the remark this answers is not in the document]
jarlebring: I agree x 2. Also: This can be considered as a part of  #70, i.e. for later.

src/NEPTypes.jl
jarlebring: Ok.

src/NEPTypes.jl
jarlebring: Done.

PR #82, opened 2018-09-14
Reexport

eringh:
Using the package `Reexport` to make things defined in the package available with only `using NonlinearEigenproblems`.
Also changes to `LOAD_PATH` when loading the package. So that the extra galleries such as `GalleryWaveguide` and `GalleryNLEVP` can be loaded with a simple `using`.


jarlebring:
Nice.

src/Gallery.jl
```diff
@@ -14,8 +14,16 @@ module Gallery
...
+    function __init__()
+        # Add the search-path to the extra galleries - so that unloaded modules can be easily loaded from top level
+        this_path = string(@__DIR__, "/gallery_extra")
+        if !(this_path in LOAD_PATH)
+            push!(LOAD_PATH, this_path)
+        end
+    end
+
+    # Add the search-path to the extra galleries - temporarily so so needed files can be included
+    push!(LOAD_PATH, string(@__DIR__, "/gallery_extra"))
```
jarlebring: Modifying the LOAD_PATH sounds a bit intrusive. Is this the recommended way to do it?

eringh: I agree it seems a bit intrusive. I have tried to look for a recommended way, or a common practice, but have been unable to find it.
Should we put the extra galleries in a separate package?
WEP can be loaded without problem since it does not require MATLAB. But not the NLEVP.

eringh: I get the feeling the semi-external modules in a package might not be a recommended practice. Or I might just be searching for the wrong type of keywords.

eringh: We should note that the marked modification of the `LOAD_PATH` does not seem to modify the load path experienced at top level (my `LOAD_PATH` in REPL is not changed when I use the package, and that statement has been there all along, I just changed the documentation of it). However, the function `__init__()` on line 17 on the other hand does this. And I think that is the part that is intrusive. If we locally change the `LOAD_PATH` while loading our package, I'm not sure that is a problem.

eringh: With that said: Does the change in directory structure really change how one would load the extra galleries? Would Julia look in a package directory to load a module that is not part of a package?

eringh: As @maxbennedich mentioned it seems possible to have mutliple packages in one repository by using top-level directories. However, that seems to require a common set of requirements (https://discourse.julialang.org/t/creating-multi-module-projects-preparing-for-pkg3/9808) and hence it would defeat the purpose of not loading GalleryNLEVP directly since we anyway need the MATLAB dependency.

Pushing the `LOAD_PATH` is mentioned in the forums (https://discourse.julialang.org/t/multiple-modules-in-single-package/5615/6) although I am not sure if that gives it credibility enough to be regarded good practice.

eringh: Another way is to keep it as it is (not push global `LOAD_PATH` in `__init__()`) but be very clear in the documentation of how to use the extra galleries. But then we need a robust way to find the extra module.

Maybe add the following code to the main module?
```
export NonlinearEigenproblems_location
function NonlinearEigenproblems_location()
    return abspath(joinpath(@__DIR__, ".."))
end
```

jarlebring: I'm not a big fan of the stdlib solution (several toml-files, less focus on what is important etc). If you want to merge now it's okay we can revisit this later. 

I think we should remember that the real source of the problem is the "using MATLAB"-dependence in GalleryNLEVP. As I mentioned to @maxbennedich, I could imagine moving GalleryNLEVP to a separate package.

maxbennedich: I googled this a bit and I haven't found any better alternatives. There's a github issue titled "Reorganizing optional package loading" with a post starting with "I found a way to lazily load a submodule on demand" which is a bit similar to our situation (perhaps you've already seen it @eringh). The suggestion is about the same as the one in this PR.

Issue #44, opened 2018-07-19
Type stability for DEPs


jarlebring:
At the meeting 2018-07-17 @eringh proposed to look into type stability. Rather than trying to make NEP-PACK type stable right now (which seems like a huge task) we can try it out for a part of the code, and see how difficult it is and if we get perforance improvement. I think we can first try to make the functions associated with one NEP-type type stable. I suggest the `DEP`.

The command `@code_warntype` is useful when investigating type stability. For instance, if you run it on a DEP to check type stability of the `size` command:
```
julia> nep=nep_gallery("dep0");
julia> @code_warntype(size(nep,1))
Variables:
  #self# <optimized out>
  nep::NEPTypes.DEP
  dim <optimized out>

Body:
  begin 
      return (Core.getfield)(nep::NEPTypes.DEP, :n)::Int64
  end::Int64
```
At the end you see that the output is Int64, so julia could determine the output type based on input type => type stable.

In contrast, if we run it on `issparse` we get
```
julia> @code_warntype(issparse(nep))
Variables:
  #self# <optimized out>
  nep::NEPTypes.DEP

Body:
  begin 
      return (NEPTypes.issparse)((Base.arrayref)((Core.getfield)(nep::NEPTypes.DEP, :A)::Array{#s739,N} where N where #s739<:(AbstractArray{T,2} where T), 1)::AbstractArray{T,2} where T)::Any
  end::Any
```
At the end we see `Any`, which means julia could not determine the output type based on the input types. I think we should try to make `issparse` and `compute_Mder()` type stable first (and then look at `compute_Mlincomb` and maybe some other functions). 

Type stable for DEP:
* [x]  `size()`
* [x]  `issparse()`
* [x]  `compute_Mder`
* [x]  `compute_Mlincomb`

[This julia discourse discussion](https://discourse.julialang.org/t/type-stability-and-array-of-matrices/12483/2) will probably be helpful. The size-function was made type-stable in ba09cd76278ba754211183c252b1bca3da9f0005 and 248f2278606cb464c0b6e409c3f34860a419c976.

eringh:
This is one of the reasons why I have been bringing up the topic if we should have the NEPs as parametric types. It is complicated, and it is true that you might have a NEP that is real but with complex eigenvalues, or vice versa. Nevertheless, I suspect it might be difficult to get some type stability in, e.g., `compute_Mder` without knowing the type of the matrices in the NEP.

[...]

PR #84, opened 2018-09-17
Configurable UMFPACK iterative refinements #72

### LINKED ISSUE #72: Support for changing number of UMFPACK iterative refinements
The number of iterative refinements in UMFPACK defaults to 2. By lowering this value, it is possible to greatly speed up the solver. For many applications, one iterative refinement might be enough (for the example below, the second refinement is not needed), while zero iterative refinements might compromise too much accuracy. Changing this value could speed up methods that rely heavily on system solves (such as NLEIGS). We could perhaps make this configurable in `DefaultLinSolver`.

```julia
julia> using SuiteSparse, LinearAlgebra, BenchmarkTools

julia> n = 10000;

julia> A = I + sprandn(n, n, 0.0005);

julia> y = randn(n);

julia> luf = lu(A);

julia> @btime global x = luf \ y; norm(A*x - y)
  52.467 ms (7 allocations: 625.20 KiB)
1.2050980166311244e-12

julia> SuiteSparse.UMFPACK.umf_ctrl[8] = 1;

julia> @btime global x = luf \ y; norm(A*x - y)
  34.722 ms (7 allocations: 625.20 KiB)
1.2050980166311244e-12

julia> SuiteSparse.UMFPACK.umf_ctrl[8] = 0;

julia> @btime global x = luf \ y; norm(A*x - y)
  15.467 ms (7 allocations: 312.70 KiB)
4.9111032477641746e-9
```

maxbennedich:
Fixes #72 

I removed the matrix `A` from `DefaultLinSolver` since it wasn't used for anything, and also removed some parametric types since as far as I could tell they didn't add anything; please let me know if I'm missing something!

### DIFF
--- a/src/LinSolvers.jl
+++ b/src/LinSolvers.jl
@@ -41,8 +41,8 @@ module LinSolvers
     """
         The linear solver associated with julia factorize()
     """
-    struct DefaultLinSolver <: LinSolver
-        Afact
+    struct DefaultLinSolver{T} <: LinSolver
+        Afact::T
         umfpack_refinements::Int
     end
 
--- a/src/LinSolvers.jl
+++ b/src/LinSolvers.jl
@@ -2,6 +2,7 @@ module LinSolvers
     using ..NEPCore
     using LinearAlgebra
     using SparseArrays
+    using SuiteSparse
     using IterativeSolvers
     using LinearMaps
     using Arpack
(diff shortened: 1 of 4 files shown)


src/LinSolvers.jl
```diff
@@ -40,29 +41,42 @@ module LinSolvers
...
     end
-    DefaultLinSolver(A::AbstractMatrix,Afact) = new(A, Afact)
 
-    function DefaultLinSolver(nep::NEP,λ)
+    function DefaultLinSolver(nep::NEP, λ, umfpack_refinements)
         A=compute_Mder(nep,λ)
         Afact=factorize(A)
-        return DefaultLinSolver{eltype(A),typeof(A)}(A, Afact)
+        return DefaultLinSolver(Afact, umfpack_refinements)
     end
```
jarlebring: I also don't directly see the use of having parametric types here. It was turned parametric at a point before we fully understood how to use parametric types well (march) 

However, in #44 we made newton + DEP type stable. With these type of changes might loose that? To be sure not to loose this along the way, we could add a unit test corresponding verifying type stability similar to #44.

jarlebring: I meant `augnewton` is type stable for DEPs: 
```
julia> dep=nep_gallery("dep0");
julia> @code_warntype(augnewton(Float64,dep))
```
So a short unit test checking could something like:
```
dep=nep_gallery("dep0");
Test.@inferred augnewton(Float64,dep,v=ones(size(dep,1)))
Test.@inferred augnewton(ComplexF64,dep,v=ones(size(dep,1)))

```

eringh: Yes, we did not fully understand how to use parametric types. However, considering the following [performance tips](https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type-1) in the Julia manual it is possible that we still want to have the types parametric. Although maybe something like
```
    struct DefaultLinSolver{T} <: LinSolver
        Afact::T
        umfpack_refinements::Int
    end

    function DefaultLinSolver(nep::NEP, λ, umfpack_refinements))
        A = compute_Mder(nep,λ)
        Afact = factorize(A)
        return DefaultLinSolver{typeof(Afact)}(Afact, umfpack_refinements))
    end
```

maxbennedich: Good suggestions. I've updated the code accordingly.

PR #86, opened 2018-09-18
Logging

maxbennedich:
Using logging framework in NLEIGS tests. We could do similarly in other unit tests as well, in order to allow customizing and disabling logging produced by tests.


jarlebring:
Looks good

PR #91, opened 2018-09-18
Fix spmf nep keyword arg

meleg:
I did not write the function  `SPMF_NEP` but I think that there was a typo in the keyword arguments. More precisely  `SPMF_NEP` should have `Schur_fact` as keyword argument. It was already like that, but with a typo.

As soon as I fixed this, other tests failed since they where written without setting properly this `Schur_fact` as keyword argument.

I have done the fix in a way that all the tests pass but I am not sure anymore this is the way we want  the function  `SPMF_NEP`, therefore I ask for a review.

As related question: do we really need `Schur_factorize_before` as field in the structure `SPMF_NEP`? If so, is there a way to avoid to pass `Schur_fact` as keyword argument to the function  `SPMF_NEP`? The fact that `Schur_factorize_before` is somehow a field of the structure `SPMF_NEP` and (with a different name) an argument of the function `SPMF_NEP`  confuses me. But I assume I did not fully understood the code.

### DIFF
--- a/src/NEPTypes.jl
+++ b/src/NEPTypes.jl
@@ -133,7 +133,7 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
  0.0  0.0
 ```
 """
-     function SPMF_NEP(AA::Vector{<:AbstractMatrix}, fii::Vector{<:Function},
+     function SPMF_NEP(AA::Vector{<:AbstractMatrix}, fii::Vector{<:Function};
             Schur_fact = false, use_sparsity_pattern = true, check_consistency=false)
 
             T=Float64;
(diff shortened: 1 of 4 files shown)


jarlebring:
It's okay to change Schur_factorize_before to a kwarg for me. Note that the previous version 
```
SPMF_NEP(nep,Av,fv,Schur_factorize_before=true)
```
made Schur_factorize_before an optional argument (not a keyword argument). Wrong choice of Schur_factorize_before matters a lot in perfornance in compute_MM. Unfortunately, I did not document exactly when it is better to use Schur_factorize_before, but please keep it.

meleg:
Ok, now I understand. Thanks for explaining. I think that it is simpler to keep it as keyword argument. In this way when we call 
`SPMF_NEP(A,f;Schur_fact=true) `
it is clear what is going on, instead with the optional argument it would look 
`SPMF_NEP(A,f,true)`
and it is not obvious what the last argument is (unless one open the code).

jarlebring:
Good. Nice work.

Issue #92, opened 2018-09-19
Not all tests are executed


jarlebring:
I think the spmf-test spmf.jl is currently never executed. Maybe there are more. If someone looks into this, maybe he can also exclude files containing # and ~, which are temporary files generated by emacs and cause me some trouble.

jarlebring:
Maybe it has something to do with my commit 08ecec4743a6e6334185b0a4c6d8210f86fbbf6b and how I used `@onlybench`.

maxbennedich:
Alright, I'll take a look.

maxbennedich:
The reason that the test was not run was that there was an `end` statement missing in the newly added test, which caused parsing of the file to fail. I fixed it by adding the `end` statement (and indented the code in the `for` block to make it more clear, but feel free to rearrange it).

While this solves this particular test, it could be useful if the test runner would actually throw an error and fail the build for invalid tests, instead of silently ignoring them. I'll see if I can update it, and also exclude temporarily generated files.

jarlebring:
Ah. Sorry about that.

PR #94, opened 2018-09-19
Add a trailing end statement in spmf test #92

maxbennedich:
We can merge this for now to fix the SPMF test, and keep #92 open for throwing an error for invalid tests.

### DIFF
--- a/test/spmf.jl
+++ b/test/spmf.jl
@@ -149,27 +149,40 @@ using SparseArrays
 
     @onlybench @testset "SPMF benchmark" begin
         # To check performance of SPMF-compute_MM function
+        large_benchmark = false
+        if large_benchmark
+            n_values = (5, 10, 300)
+            p_mm_values = (5, 10, 15, 100, 1000)
+            p_mder_values = (1, 3, 10, 20)
+            m_terms = 100
+        else
+            n_values = (5, 10)
+            p_mm_values = (5, 10, 15, 100)
+            p_mder_values = (1, 3, 10)
+            m_terms = 20
+        end
+
         for Schur_fact in (true, false)
-            for n in (5,10,300)
+            for n in n_values
                 # small number of terms.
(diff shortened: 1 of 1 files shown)


maxbennedich:
Hmm, the benchmark run failed with:
```
Running test spmf (29 / 45)
No output has been received in the last 10m0s, this potentially indicates a stalled build ...
```
@jarlebring Do you know how long the new benchmark test that you added should take to run? Could it take longer than 10 minutes?

eringh:
Last time we had that type of failure in a test it was because of a stack overflow which was never triggered. At that time we speculated if it was the Julia interpreter that broke down before throwing an error. Then adding a print in the function triggered a stack overflow.

What happens when you run the tests locally?

jarlebring:
It takes 5 sec on my computer. It does require some memory.

maxbennedich:
I think it is the size of the problem that causes the timeout. When I run the `@onlybench` test locally, the code within the `for n in (5,10,300)` loop takes around 5 seconds for `n = 5`, 10 seconds for `n = 10`, and 70 seconds for `n = 30`. I didn't test higher than that, but it seems reasonable that `n = 300` could take longer than 10 minutes.

Perhaps we should remove `n = 300` and just test `n = 5` and `n = 10`? Even with that limitation, it will still be the slowest test among all the benchmarks.

Alternatively, we could add another level of benchmarks, like `@fullbench`, that are not run as part of the build and the automated benchmark suite, but run manually on demand.

jarlebring:
I would say we drastically decrease the numbers for now. When I do benchmarking (e.g. locally) I can change it back.

maxbennedich:
Alright, I updated the test, and added a flag in the test to easily switch the parameters to run a larger benchmark.

maxbennedich:
Merging this for now so that the test will pass once I fix the remaining issue in #92.

PR #102, opened 2018-09-21
SPMF rewrite part 1

jarlebring:
This is a quite big rewrite of the SPMFs. 

* Separation of SPMF_NEP into SPMF_dense and SPMF_sparse. From a user-perspective nothing changed. It makes @maxbennedich sparsity pattern optimization more separated from the rest.
* compute_MM rewrite, with type logic. The output of compute_MM is now type stable (but there are some red Any in `code_warntype` which I don't know if they matter).
* compute_Mder now has two different functions for SPMF_dense and SPMF_sparse 
* The SPMFs are now parametric both wrt to the T:matrices, but also Ftype. Ftype defines a type logic for the f-functions. More precisely  `eltype(F(λ))=promote_type(eltype(λ),Ftype)`. That way the output type can inferred from the input types of  compute_MM. 

I thought making it type stable would make a much bigger difference. I did find a (somewhat strange) example where it makes a difference.  Standalone illustration between current master (91fdc26ca972d60c67ae2b6f2a433dc44f160b42) and this PR:
```julia
    n=100; m=10000; p=5; Schur_fact=false;
    fv=Vector{Function}(undef,m);
    Av=Vector{Matrix{Float64}}(undef,m);
    for k=1:m
        if (mod(k,2)==1)
            fv[k] = S-> inv((1im+sqrt(k))*one(S)-S); # 1/((i+√k)I-S)
        else
            fv[k] = S-> inv(sqrt(k)*one(S)-S); # 1/(I√k-S)
        end
        Av[k] = randn(n,n)
    end
    V=randn(n,p)+1im*randn(n,p);
    S=randn(p,p);
    spmf2=SPMF_NEP(Av,fv,Schur_fact = Schur_fact,check_consistency=false);
    @btime compute_MM(spmf2,S,V);
```
On master: 
```
  742.639 ms (248469 allocations: 292.95 MiB)
```
This PR:
```
  688.245 ms (205003 allocations: 214.63 MiB)
```

I'm not quite finished with this. I also want to implement `compute_Mder` for higher derivatives, and double check that there was no regression in the sparse summation code.

### DIFF
--- a/src/NEPTypes.jl
+++ b/src/NEPTypes.jl
@@ -125,8 +125,8 @@ for matrices in the standard matrix function sense.
     end
     SPMF_NEP{T} = Union{SPMF_NEP_dense{T},SPMF_NEP_sparse{T}}
 
-    SPMF_NEP(n, A, fi, Schur_factorize_before, Zero) =
-        SPMF_NEP(n, A, fi, Schur_factorize_before, Zero, Vector{SparseMatrixCSC{Float64,Int}}())
+    #SPMF_NEP(n, A, fi, Schur_factorize_before, Zero) =
+    #    SPMF_NEP(n, A, fi, Schur_factorize_before, Zero, Vector{SparseMatrixCSC{Float64,Int}}())
 
 """
      SPMF_NEP(AA, fii, Schur_fact = false, use_sparsity_pattern = true, check_consistency=true)
@@ -168,7 +168,7 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
          end
 
          if (size(AA,1)==0)
-             return SPMF_NEP{typeof(I)}(0); # Create empty SPMF_NEP.
+             return SPMF_NEP(0); # Create empty SPMF_NEP.
          end
          n=size(AA[1],1);
 
@@ -184,42 +184,48 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
          end
 
 
-         T = eltype(AA[1])
-         As = Vector{SparseMatrixCSC{T,Int}}()
 
-
-         if use_sparsity_pattern && issparse(AA[1])
-             # Merge the sparsity pattern of all matrices without dropping any zeros
-             Zero = LinearAlgebra.fillstored!(copy(AA[1]), 1)
-             for i = 2:size(AA,1)
-                 Zero += LinearAlgebra.fillstored!(copy(AA[i]), 1)
-             end
-             Zero = T.(Zero)
-             Zero.nzval[:] .= T(0)
-
-             # Create a copy of each matrix with the sparsity pattern of all matrices combined
-             @inbounds for A in AA
-                 S = copy(Zero)
-
-                 for col = 1:size(A, 2)
-                    for j in nzrange(A, col)
-                        S[A.rowval[j], col] = A.nzval[j]
-                    end
-                 end
-
-                 push!(As, S)
-             end
+         if !(eltype(AA) <: SparseMatrixCSC)
+             # Dense
+             this=SPMF_NEP_dense{typeof(AA[1])}(n,AA,fii,Schur_fact);
          else
-             Zero=zeros(n,n)
+             # Sparse: Potentially do the joint sparsity pattern trick.
+             T = eltype(AA[1])
+             As = Vector{SparseMatrixCSC{T,Int}}()
+
+             if use_sparsity_pattern && issparse(AA[1])
+                 # Merge the sparsity pattern of all matrices without dropping any zeros
+                 Zero = LinearAlgebra.fillstored!(copy(AA[1]), 1)
+                 for i = 2:size(AA,1)
+                     Zero += LinearAlgebra.fillstored!(copy(AA[i]), 1)
+                 end
+                 Zero = T.(Zero)
+                 Zero.nzval[:] .= T(0)
+
+                 # Create a copy of each matrix with the sparsity pattern of all matrices combined
+                 @inbounds for A in AA
+                     S = copy(Zero)
+
+                     for col = 1:size(A, 2)
+                         for j in nzrange(A, col)
+                             S[A.rowval[j], col] = A.nzval[j]
+                         end
+                     end
+
+                     push!(As, S)
+                 end
+             else
+                 Zero=zeros(n,n)
+             end
+
+             this=SPMF_NEP_sparse{typeof(AA[1])}(n,AA,fii,Schur_fact,Zero,As);
          end
 
-
-         this=SPMF_NEP_sparse{typeof(AA[1])}(n,AA,fii,Schur_fact,Zero,As);
          return this
     end
     function SPMF_NEP(n) # Create an empty NEP of size n x n
-         Z=zeros(n,n)
-         return SPMF_NEP(n,Vector{Matrix}(),Vector{Function}(),false,Z);
+        Z=zeros(n,n)
+        return SPMF_NEP_dense{AbstractMatrix}(n,Vector{Matrix}(),Vector{Function}(),false);
     end
     function compute_MM(nep::SPMF_NEP,S::AbstractMatrix,V::AbstractVector)
         if (issparse(V))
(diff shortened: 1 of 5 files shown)


jarlebring:
My general observation was that it seems difficult to have one piece of code which is always the "best" for `compute_MM(nep,S,V)`. Depending on the size of the matrices `S` and `V` and `size(nep)` there are different "best" choices.

src/NEPTypes.jl
```diff
@@ -102,19 +111,31 @@ All of the matrices ``A_0,...`` are of size ``n×n``
 and ``f_i`` are a functions. The  functions ``f_i`` must be defined
 for matrices in the standard matrix function sense.
 """
-    struct SPMF_NEP <: AbstractSPMF
+
+# Logic behind Ftype:
+#  The eltype(F(λ))=promote_type(eltype(λ),Ftype)
+
+    struct SPMF_NEP_dense{T,Ftype}  <: AbstractSPMF{T}  # T should be a AbstractMatrix
```
maxbennedich: If `T` has to be an `AbstractMatrix`, perhaps we could do `struct SPMF_NEP_dense{FType, T<:AbstractMatrix}`?

src/NEPTypes.jl
```diff
@@ -179,122 +193,121 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
              end
          end
 
+         if !(eltype(AA) <: SparseMatrixCSC)
+             # Dense
+             this=SPMF_NEP_dense{typeof(AA[1]),Ftype}(n,AA,fii,Schur_fact);
```
maxbennedich: If we create the following constructor:
```julia
SPMF_NEP_dense{FType}(n, A, fi, Schur_fact) where FType =
    SPMF_NEP_dense{FType, eltype(A)}(n, A, fi, Schur_fact)
```
we should be able to skip the redundant type:
```julia
this = SPMF_NEP_dense{Ftype}(n, AA, fii, Schur_fact)
```
(at the cost of maintaining an extra constructor)

src/NEPTypes.jl
```diff
@@ -179,122 +193,121 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
...
+         end
 
-         this=SPMF_NEP(n,AA,fii,Schur_fact,Zero,As);
          return this
     end
     function SPMF_NEP(n) # Create an empty NEP of size n x n
-         Z=zeros(n,n)
-         return SPMF_NEP(n,Vector{Matrix}(),Vector{Function}(),false,Z);
+        Z=zeros(n,n)
+        return SPMF_NEP_dense{AbstractMatrix}(n,Vector{Matrix}(),Vector{Function}(),false);
```
maxbennedich: Does this work, with a single type like this?

src/NEPTypes.jl
```diff
@@ -179,122 +193,121 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
...
+        VFi=Matrix{promote_type(FStype,eltype(V))}(undef,n,p); # Type of V*f(S)
+        local Fi=Matrix{FStype}(undef,p,p);
+        for i=1:m
             ## Compute Fi=f_i(S) in an optimized way
             if (isdiag(S)) # optimize if S is diagonal
                 Sd=diag(S);
+                local Fid::Vector{FStype}
                 if (norm(Sd .- Sd[1])==0) # Optimize further if S is a multiple of identity
-                    Fid=nep.fi[i](reshape([Sd[1]],1,1))[1]*ones(size(Sd,1))
+                    Fid=nep.fi[i](Sd[1])*ones(T0,p)
```
maxbennedich: You may be able to use `fill` for this.

src/NEPTypes.jl
```diff
@@ -179,122 +193,121 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
...
-            ## Sum it up
-
-
-            VFi=V*Fi;
-            if (isa(nep.A[i],SubArray) && (eltype(nep.A[i]) != eltype(VFi)))
-                # 2018-09-14: SubArray x Matrix of different types do not work? Check in new version of Julia/base at later point (TODO)
-                # https://discourse.julialang.org/t/subarray-x-matrix-multipliciation-of-different-eltype-fails/14950
-                # Workaround by making a matrix copy.
-                Z=Z+copy(nep.A[i])*VFi;
+                Fi[:,:]=Diagonal(Fid)
```
maxbennedich: Could we use fusing dot here instead of `[:,:]`? I.e. `Fi .= Diagonal(Fid)`. And similarly below. A quick benchmark suggests that fusing dot is a lot faster, but I'm not so familiar with the colon syntax, perhaps it does something else?

src/NEPTypes.jl
```diff
@@ -179,122 +193,121 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
...
-                                convert.(TZ, nep.As[1].nzval .* x[1]))
-            for k = 2:length(nep.As)
-                Z.nzval .+= nep.As[k].nzval .* x[k]
-            end
+     function compute_Mder(nep::SPMF_NEP_dense,λ::Number)
+         # figure out the return type, as the greatest type of all input
+         x = map(i -> nep.fi[i](reshape([λ],1,1))[1], 1:length(nep.fi))
+         Tx = mapreduce(eltype, promote_type, x)
+         TA=mapreduce(eltype, promote_type, nep.A); # Greatest type of all A-matrices
+         TZ=promote_type(TA,Tx)  # output type
```
maxbennedich: Maybe we could extract the 5 lines above here, that are repeated below, to its own function? That would make it easier and safer to refactor that code in the future.

src/NEPTypes.jl
```diff
@@ -1016,38 +1030,41 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
+        Fλtype=promote_type(eltype(λ),Ftype);
+        TT=promote_type(Fλtype,eltype(V)); # Return type
+
+        z=zeros(TT,n)
+        for i=1:size(nep.A,1)
+            # Get the function value if V is a vector,
+            # otherwise get a vector of scaled derivatives
+            Fi1=(V isa AbstractVector) ? nep.fi[i](S) : nep.fi[i](S)[:,1]
+            VFi1=V*Fi1
+            z[:] += nep.A[i]*VFi1
```
maxbennedich: This could perhaps also use fusing dot?

jarlebring:
In several places in the code we have sums like this:
```julia
nep=nep_gallery("qdep0");
n=size(nep,1); S=randn(3,3);; V=randn(n,3); 
z=zeros(ComplexF64,n);
for i=1:size(nep.A,1)
   Fi1=nep.fi[i](S)[:,1];
   VFi1=V*Fi1
   z[:] += nep.A[i]*VFi1
end
```
This can also be done with a single `mapreduce` call:
```
z2=mapreduce(jj -> nep.A[jj]*(V*nep.fi[jj](S)[:,1]), +, [1:size(nep.A,1);])
```
I'm not so experienced with mapreduce. Does someone know right away if this is advisable?

src/NEPTypes.jl
```diff
@@ -1016,38 +1030,41 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
+            S=λ # Vector means just compute matrix vector
         else
-            for i=1:size(nep.A,1)
-                Fi=nep.fi[i](S)[:,1];
-                z=z .+ nep.A[i]*(V*Fi);
-            end
-    	end
+            # V matrix means compute linear combination of derivatives. Use
+            # scaling trick
+       	    S=diagm(0 => λ*ones(eltype(λ),k)) + diagm(-1 => (a[2:k]./a[1:k-1]).*(1:k-1))
```
meleg: Very marginal comment:
should you define `TT ` earlier and then `S` with the correct type? Namely

```
        # Type logic
        Fλtype=promote_type(eltype(λ),Ftype);
        TT=promote_type(Fλtype,eltype(V)); # Return type
     	S=diagm(0 => λ*ones(TT,k) + diagm(-1 => (a[2:k]./a[1:k-1]).*(1:k-1)) 
```
Or at least, when you define `S` has to have the same type of `a`
`        	    S=diagm(0 => λ*ones(eltype(a),k)) + diagm(-1 => (a[2:k]./a[1:k-1]).*(1:k-1)) `
It looks like a good practice to compute the type at the very top of the function.

src/NEPTypes.jl
jarlebring: You're right. Updated. I kept order of the params.

src/NEPTypes.jl
jarlebring: You're right. Fixed.

src/NEPTypes.jl
jarlebring: Ideally, users would not construct SPMF_NEP_dense themselves, but rely on `SPMF_NEP()`. Your suggestion would encourage direct creation, so I would prefer not to implement this, although the code gets slightly shorter.

src/NEPTypes.jl
jarlebring: Correct.

src/NEPTypes.jl
jarlebring: I agree.

src/NEPTypes.jl
jarlebring: Good idea. Done.

src/NEPTypes.jl
jarlebring: Indeed. It should be precomputed, and it is currently not correct.

jarlebring: Fixed.

src/NEPTypes.jl
```diff
@@ -102,19 +111,31 @@ All of the matrices ``A_0,...`` are of size ``n×n``
...
+#  The eltype(F(λ))=promote_type(eltype(λ),Ftype)
+
+    struct SPMF_NEP_dense{T<:AbstractMatrix,Ftype}  <: AbstractSPMF{T}  # T should be a AbstractMatrix
+        n::Int
+        A::Vector{T}   # Array of Array of matrices
+        fi::Vector{Function}  # Array of functions
+        Schur_factorize_before::Bool # Tells if you want to do the Schur-factorization
+    end
+
+    struct SPMF_NEP_sparse{T,Ftype}  <: AbstractSPMF{T}  # T should be a AbstractMatrix
```
maxbennedich: Here we could also do `T<:AbstractMatrix`, right?

jarlebring: Yep.

jarlebring:
This became such a big PR. I think I will merge this now and leave the mder-optimization and work on joint sparsity pattern to a different pr.

PR #107, opened 2018-09-22
SPMF rewrite2: compute_Mder + sparsity alignment

jarlebring:
Here is the second round of SPMF rewrite:

* After moving the sparsity alignment trick to the constructor, I was able to remove the need for `As` in `SPMF_NEP_sparse`. This lead to the fact that SPMF_dense and SPMF_sparse were so similar that they could be merged to one type again. Quite a bit of cleanup work. Not so much efficiency issues.  
* I tried to factorize the code into helper functions, since I believe the same problems will later need to be solved for more sparse AbstractSPMF, e.g., DEP{SparseMatrixCSC}.  
* I wrote a specialized version of `compute_Mder(::SPMF_NEP`. This lead to a huge improvement for higher derivatives, mostly due to how slow compute_Mder_from_MM is when size(nep) is large:

```julia
julia> nep=nep_gallery("qdep0");
julia> @btime Z=compute_Mder_from_MM(nep,λ,1); # Old
  4.377 s (158 allocations: 1.27 GiB)
julia> @btime Z=compute_Mder(nep,λ,1); # New
  168.745 μs (138 allocations: 818.25 KiB)
```
The only method I know using derivative matrices (and not compute_Mlincomb) is mslp, so there will be some benefits. 

@maxbennedich: I moved around the sparsity alignment code. The sparsity alignment was very important for you in the nleigs-tests. Do you think you could see that the benefits are not lost? Or tell me how to check it.

### DIFF
--- a/src/NEPTypes.jl
+++ b/src/NEPTypes.jl
@@ -13,7 +13,7 @@ module NEPTypes
     export REP
     export SPMF_NEP
     export SPMF_NEP_dense
-    export SPMF_NEP_sparse
+
     export AbstractSPMF
     export SumNEP, SPMFSumNEP, GenericSumNEP
     export Proj_NEP;
@@ -312,7 +312,7 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
          TZ,x = compute_Mder_fi_and_output_type(nep,λ)
          if (nep.sparsity_patterns_aligned)
              # Sparsity patterns are aligned, so just change
-             # the value entires in the sparse matrix object
+             # the value entries in the sparse matrix object
              Z = SparseMatrixCSC(nep.A[1].m, nep.A[1].n,
                                  copy(nep.A[1].colptr), copy(nep.A[1].rowval),
                                  copy(convert.(TZ, nep.A[1].nzval .* x[1])))
@@ -329,13 +329,31 @@ julia> compute_Mder(nep,1)-(A0+A1*exp(1))
          return Z
     end
 
-    function compute_Mder(nep::SPMF_NEP,λ::Number,i::Integer)
+    # For higher derivatives
+    function compute_Mder(nep::SPMF_NEP{T,Ftype},λ::Number,i::Integer) where {T,Ftype}
         if (i==0)
             return compute_Mder(nep,λ);
         else
-            # This is typically slow for i>1 (can be optimized by
-            # treating the SPMF-terms individually)
-            return compute_Mder_from_MM(nep,λ,i)
+
+            local n,k;
+            n=size(nep,1);
+            k=i+1;
+
+       	    S=diagm(0 => fill(λ,k), -1 => (1:k-1)) # Jordan matrix trick
+            TS=eltype(S);
+
+            # Type logic
+            Fλtype=promote_type(TS,Ftype);
+            TT=promote_type(Fλtype,eltype(nep.A[1])); # Return type
+
+            z=zeros(TT,n)
+            # No alignment of sparsity pattern exploitation implemented. Naive summing.
+            Z::SparseMatrixCSC{TT,Int} = nep.A[1]*nep.fi[1](S)[end,1];
+            for j = 2:length(nep.A)
+                Z .+= nep.A[j] * nep.fi[j](S)[end,1];;
+            end
+    	    return Z
+
         end
     end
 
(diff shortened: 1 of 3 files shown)


maxbennedich:
Impressive work!

I ran a benchmark for the NLEIGS tests only, and there does indeed seem to be a regression in the "NLEIGS: Particle variant R2" test (in `nleigs_particle_variant_r2.jl`). Not sure yet what caused it, but this is the problem that has 81 nonlinear matrices, so it seems likely that the regression is related to the sparsity alignment trick or SPMF addition.

See benchmark details below. This is in my experimental branch only for now, I'll merge it to master soon and post some instructions on how to do the benchmarks.

image

maxbennedich:
The problem seems to be that `align_sparsity_patterns` defaults to `false` in this branch, while it used to default to `true`. Is that intentional?

jarlebring:
Ah. Yes. Thanks. I changed it in the nleigs-test and now I get lower memory usage. 

Default value of joint pattern = false is intentional. The computation of joint sparsity pattern is very memory economical: all the sparse matrices point to the same colptr and rowval vectors (yes really "the same" not copies).  As a consequence, modifying one matrix will normally change all matrices, i.e. the joint sparsity pattern matrices should be viewed as read only. This seemed like a bad thing to enforce by default.

src/NEPTypes.jl
```diff
@@ -115,34 +115,27 @@ for matrices in the standard matrix function sense.
...
 
-    #SPMF_NEP(n, A, fi, Schur_factorize_before, Zero) =
-    #    SPMF_NEP(n, A, fi, Schur_factorize_before, Zero, Vector{SparseMatrixCSC{Float64,Int}}())
+    # Alias for SPMFs which contain sparse matrices
+    SPMF_NEP_sparse{T<:AbstractSparseMatrix,Ftype}=SPMF_NEP{T,Ftype}
+
 
 """
-     SPMF_NEP(AA, fii, Schur_fact = false, use_sparsity_pattern = true, check_consistency=true)
+     SPMF_NEP(AA, fii, Schur_fact = false, join_sparsity_patterns = false, check_consistency=true)
```
maxbennedich: Minor: this should be "align_sparsity_patterns" right?

maxbennedich:
Ok, good! I'll run the comparison again, I can run it on all tests as well just to make sure nothing else was affected.

I agree it's quite precarious to share the colptr/rowval vectors between matrices, but I'm also not sure if a `NEP` instance (including `A` matrices) should ever be modified after it's created? It feels safer to me to always consider them read-only after creation, and create a new instance if we need to modify one. But perhaps that's not reasonable in practice?

jarlebring:
If we find a way to enforce the sparse matrices to be constant, I would be in favour of changing default value. Is it possible to make colptr/rowval immutable? I found packages introducing static vectors, but not immutable vectors.

maxbennedich:
Hmm, I don't think that's doable with the default sparse matrix, since the colptr/rowval types are concretely defined as Vectors.

I ran the benchmarks now and everything seems good!

maxbennedich:
However, the test `DEP WITH GENERIC Y0` in `iar_chebyshev.jl` seems to have gotten over 2x slower at some point during or around the first SPMF rewrite. Any idea what might have caused this? Otherwise, I'll look into it a bit, and create a new issue if needed. (It doesn't have to be a regression of course, perhaps the test itself changed.)

Before (test altered to only run `DEP WITH GENERIC Y0`):
```julia
julia> include("test/iar_chebyshev.jl")
  1.307680 seconds (47.90 k allocations: 520.383 MiB, 14.66% gc time)
```

After:
```julia
julia> include("test/iar_chebyshev.jl")
  2.840238 seconds (52.21 k allocations: 497.758 MiB, 6.47% gc time)
```

maxbennedich:
Ah, that seems to have been caused by the NEP used for that test being a mix of dense and sparse matrices:
```julia
A0 = sparse(K, J, rand(3*n-2))
A1 = sparse(K, J, rand(3*n-2))
nep = SPMF_NEP([Matrix(1.0I, n, n), A0, A1], [λ -> ...
```
It seems that we used to support this, but after the SPMF rewrite, all the matrices will be converted to dense as long as any matrix is dense (?). Changing the test like this makes it 6 times faster:
```julia
nep = SPMF_NEP([sparse(1.0I, n, n), A0, A1], [λ -> ...
```
Should we no longer support a mix of dense and full matrices? If so, perhaps we should throw an error or warning if that is passed as input, to not give a user the impression that we support that?

maxbennedich:
Test updated here for now: https://github.com/nep-pack/NonlinearEigenproblems.jl/commit/0ccf7fb35b6ac50d7bbf65b1a7e38d5b6ede3f1e

jarlebring:
Yes. You are right. Mixing matrix types is not allowed within one SPMF_NEP. Thanks for looking into this.

PR #108, opened 2018-09-24
Iar chebyshev test should not mix dense and sparse matrices

maxbennedich:
Cf. discussion in #107


jarlebring:
:+1:

maxbennedich:
I added a check to enforce that the matrices are either all sparse or all dense, and also ran into some trouble and created a new issue in #114.

maxbennedich:
I also did #114 as part of this PR.

jarlebring:
:+1:

Issue #104, opened 2018-09-21
Clean up compute_Mlincomb(nep::DEP


jarlebring:
The compute_Mlincomb(::DEP can be cleaned up a bit:

* should be compute_Mlincomb!
* fusing dots
* type logic does not take `a` into account. 
* TODO:doc- tag?
* the if k>1 should not be necessary considering how it is handled in other compute_mlincombs

meleg:
After the fix on the function compute_Mlincomb we get the following improvement for the instance:
```
using NonlinearEigenproblems
using Test
using LinearAlgebra
using IterativeSolvers

nep=nep_gallery("dep0_tridiag",500000)
n=size(nep,1);	k=50;
V=rand(n,k);	λ=rand()*im+rand(); a=rand(k)
z1=compute_Mlincomb!(nep,λ,copy(V),a)
compute_Mlincomb(nep,λ,V,a)
@time z1=compute_Mlincomb!(nep,λ,V,a)
```
On the branch compute_Mlincomb_DEP
`  0.164676 seconds (33 allocations: 41.967 MiB, 32.51% gc time)`
on the master
`  0.211399 seconds (56 allocations: 118.263 MiB, 32.24% gc time)`

PR #109, opened 2018-09-24
Compute mlincomb dep

meleg:
I have changed compute_Mlincomb for DEP by following the directions in the Issue #104.


src/NEPTypes.jl
```diff
@@ -1004,33 +1004,40 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
         end
-        if k>1 z[:]-=view(V,:,2:2) end
-        z[:]-=λ*view(V,:,1:1);
-        return z
+
+        if (V isa AbstractVector) || k == 1
+            z -= λ*V;
+        else
+            z -= view(V,:,2:2)
+            z .-= λ*view(V,:,1:1);
```
jarlebring: Why working with views here? Also: Can't this be done more efficiently with one call to `muladd`?

src/NEPTypes.jl
```diff
@@ -1004,33 +1004,40 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
+        end
+        return z[:]
```
jarlebring: Why do you have a `[:]`? If its because `z` turns into a n x 1 matrix at some point, I think it is better to fix that.

src/NEPTypes.jl
```diff
@@ -1004,33 +1004,40 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
         n=size(V,1); k=size(V,2);
         Av=get_Av(nep)
         # determine type froom greates type of (eltype probtype and λ)
-        T=promote_type(promote_type(eltype(V),typeof(λ)),eltype(Av[1]),eltype(nep.tauv))
-        if (k>1)
-            broadcast!(*,V,V,transpose(a))
+        TT=promote_type(promote_type(eltype(V),typeof(λ)),eltype(Av[1]),eltype(nep.tauv),eltype(a))
+
+        if (V isa AbstractVector)
+            V=a[1]*V;
```
jarlebring: More efficient?  `rmul!(V,a[1])`?

src/NEPTypes.jl
meleg: I did not know the function `muladd`. I will fix it.

jarlebring:
The "todo" can also be removed. A couple of code comments would be nice.

src/NEPTypes.jl
meleg: Yes! For a sparse DEP of size 5 millions we have:
Old version 
  `0.575768 seconds (37 allocations: 648.500 MiB, 42.12% gc time)`
New version
`  0.489866 seconds (32 allocations: 495.912 MiB, 35.11% gc time)`
(code comes in next commit)

src/NEPTypes.jl
```diff
@@ -1042,33 +1042,39 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
-                z[:]+=Av[j+1]*(V*w[1]);
-            end
+            w=Array{TT,1}(exp(-λ*nep.tauv[j])*(-nep.tauv[j]) .^(0:k-1))
+            mul!(Vw, V, w)
+            mul!(AVw, Av[j+1], Vw)
+            z[:] .+= AVw;
+        end
+
+        if (V isa AbstractVector) || k == 1
+            z-=rmul!(V,λ)
```
jarlebring: I would use a fuse subtrastraction ` .-=`, for efficiency and in order to not potentially change the type of `z`. The case `V` vector and matrix need to be handled differently? Or maybe with `V[:]`?

meleg:
Ok, it seems now it works properly and I implemented all the suggestions. Can I merge?

jarlebring:
I'm going to be an annoying perfectionist now (because I think this will be a show-case function of how it can be done perfectly). Isn't it possible to write the function with modifying `V`? The for-loop can incorporate a into the w-vector easily and the final substractions can just scale with `a[1]` and `a[2]` and remove the `!` in rmuls.

jarlebring:
Another slight reduction of allocations: Rather than calling `get_Av(::DEP)`, access the `nep.A` directly. Note that `get_Av` allocates a sparse identity matrix.

src/NEPTypes.jl
```diff
@@ -1041,34 +1041,39 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
+        end
+
+        # distinguis the case V is a vector and V is a matrix
+        # fix with the proper derivative count
+        if (V isa AbstractVector)
+            z[:] .-= a[1]*λ*V
+        elseif k==1
+            z[:] .-= a[1]*λ*V[:]
+        else
+            z .+= muladd(-λ,a[1]*V[:,1],-a[2]*V[:,2])
```
jarlebring: `muladd(-λ*a[1],V[:,1],-a[2]*V[:,2])` ?

src/NEPTypes.jl
```diff
@@ -1041,34 +1041,39 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
+            mul!(AVw, nep.A[j], Vw)
+            z[:] .+= AVw;
+        end
+
+        # distinguis the case V is a vector and V is a matrix
+        # fix with the proper derivative count
+        if (V isa AbstractVector)
+            z[:] .-= a[1]*λ*V
+        elseif k==1
+            z[:] .-= a[1]*λ*V[:]
```
jarlebring: Two changes: `z .-= (a[1]*λ)*V[:]`  also above.

jarlebring:
Good job :+1:. Ready to be merged from my perspective.

PR #111, opened 2018-09-25
Expand projection matrices

jarlebring:
This provides the possibility to expand projected problems with `expand_projectmatrices!` instead of always setting it. I also updated `nlar`. 

Originally, I thought the function would be 
```expand_projectmatrices!(nep,w::Vector,v::Vector).```
In the end I ended up with 
```expand_projectmatrices!(nep,Wnew::AbstractMatrix,Vnew::AbstractMatrix)```
where `Wnew` and `Vnew` are the complete subspace matrices. This was in order to avoid having a copy of the `V` and `W` matrices in the `pnep`. 

Some things (like documentation update) still remain. I will wait for other nlar-related PRs before merging.

### CODE REVIEW

File: src/NEPTypes.jl [hunk 1]
```diff
@@ -893,22 +884,45 @@ julia> compute_Mder(nep,3.0)[1:2,1:2]
...
-        B1=view(Matrix{T}(copy(W')*nep.orgnep_Av[1]*V),1:k,1:k);
-        T_sub = typeof(B1)
-        # The coeff matrices for the SPMF_NEP created in the end
+        @assert(k <= size(nep.projnep_B_mem[1],1))
+        # Reliable way to determine SubArray-like type
+        T_sub = typeof(view(nep.projnep_B_mem[1],1:1,1:1))
         B = Vector{T_sub}(undef,m);
-        B[1]=B1;
-        for i=2:m # From 2 since we already computed the first above
+        for i=1:m # From 2 since we already computed the first above
```
maxbennedich: Minor: the comment is outdated now

src/NEPTypes.jl
```diff
@@ -893,22 +884,45 @@ julia> compute_Mder(nep,3.0)[1:2,1:2]
...
+        # Reliable way to determine SubArray-like type
+        T_sub = typeof(view(nep.projnep_B_mem[1],1:1,1:1))
+        B = Vector{T_sub}(undef,m);
+        @assert(k+1 <= size(nep.projnep_B_mem[1],1))
+        for i=1:m
+            # Expand the B-matrices
+            nep.projnep_B_mem[i][1:k,k+1]=copy(Wnew[:,1:k]')*Av[i]*v;
+            nep.projnep_B_mem[i][k+1,1:(k+1)]=w'*Av[i]*Vnew[:,1:(k+1)];
+            B[i]=view(nep.projnep_B_mem[i],1:(k+1),1:(k+1));
+        end
```
maxbennedich: Not sure if you'd find this more or less readable, but one way to get rid of the `T_sub` logic would be to use `map`, something like this:
```julia
B = map(i -> begin
        # Expand the B-matrices
        nep.projnep_B_mem[i][1:k,k+1]=copy(Wnew[:,1:k]')*Av[i]*v;
        nep.projnep_B_mem[i][k+1,1:(k+1)]=w'*Av[i]*Vnew[:,1:(k+1)];
        view(nep.projnep_B_mem[i],1:(k+1),1:(k+1));
    end, 1:m)
```
Also, in case we need to think about performance here, it feels like some of those factors could be moved out of the loop and/or the indexing in `Vnew` and `Wnew` could be done as views.

jarlebring: This can certainly be moved outside the loop: `copy(Wnew[:,1:k]')`. The matrices Vnew and Wnew are commonly already SubArray (e.g. the way they are used in nlar). Can you be more specific about performance? I dont see what can be factorized.

maxbennedich: I was thinking of `Wnew` and `Vnew`. I think indexing will create a copy even if they're already SubArrays? In that case `view` would save some allocations. Is the `copy` needed for the transposed multiplication to work?

jarlebring: Ok. I will fix it. The copy(V') is necessary since it is otherwise very slow for dense matrices (similar to the example you showed).

jarlebring:
I updated `nlar`, but not `iar` or `jd` which also use projected problems. @eringh and @meleg: If you want more efficiency you can update those to use `expand_projectmatrices!`.

PR #112, opened 2018-09-25
Benchmarks and regression analysis

maxbennedich:
I think the benchmark / regression analysis branch is ready to be merged now. I've described how it works in more detail [on the wiki](https://github.com/nep-pack/NonlinearEigenproblems.jl/wiki/Benchmarks-and-regression-analysis).

Currently, Travis uploads the reports to a repo in my name. If you think this is useful, we can instead put this in a "julia-ci" repository in nep-pack. The idea is that Travis generates and stores a JSON file for each run, which we can use to analyze CPU and memory usage for recent commits. Unfortunately, the numbers are quite unreliable due to the different CPU platforms Travis uses, but I still think it can be useful. (It'd also be a good idea to work on making our tests more stable; both an `iar` and a `tiar` test currently fail for me on `eight` for example.)


jarlebring:
Wow. Cool. 

I have the same problem on eight. The tests for iar and tiar still seem lead to different round-off errors and fail, although they seem to actually work. @meleg can you look into running them on eight?

meleg:
@jarlebring I will try to work on it. I have a lot of teaching in this week, so the process can be slow.
It will help if you provide me very specific instances.

maxbennedich:
If there are no objections, I'll create a `julia-ci` repository in `nep-pack` to host the JSON benchmark reports.

jarlebring:
sounds good

PR #121, opened 2018-09-28
Make testing / benchmarking its own project

### LINKED ISSUE #119: Make testing / benchmarking its own project
The test and benchmarking code has many dependencies that are not needed by our core package. Currently, all of those dependencies are present in the main `Project.toml` file. A convenient way to separate the dependencies would be to make this code its own project. Basically, it's done by creating a new `Project.toml` file in the `test` directory, and moving the test-specific dependencies there, plus a reference to the parent project by path `".."`.

With this method, we could also support more than two environments. For example if we have some plotting code, or MATLAB dependent code, we could make those their own projects with custom `Project.toml` files.

Some more details in [this discourse post](https://discourse.julialang.org/t/psa-use-a-project-for-building-your-docs/14974/1) (it talks about documentation, but the same idea could be applied to any subproject).

maxbennedich:
Fixes #119 

Benefits:

- Test- and benchmark-specific dependencies moved out of the main `Project.toml` and into its own project in `test/`
- Test code then gets its own top-level module, which I named `NonlinearEigenproblemsTest`, meaning that we don't need the `LOAD_PATH` hack anymore

Drawbacks:

- We have two environments to keep track of; the main project and the test project. When running test code, the test project must be specified. This can be done from the command line like this:
```julia
$ julia --project=test test/serialization.jl
```
Alternatively:
```julia
$ julia --project=test
...
julia> include("test/serialization.jl")
```
Or from Pkg REPL-mode in Julia:
```julia
(v1.0) pkg> activate test

julia> include("test/serialization.jl")
```
- Similarly, when adding new dependencies, we need to make sure to add them to the correct project. (We already need to do this with the global project vs the package project.)


jarlebring:
I guess this is the only sensible solution since we want considerably more packages in the tests / benchmarks. Does `Base.runtests()` work as usual?

maxbennedich:
It's the best I could come up with that removes the dependencies from the root project. I think better solutions will be available in future versions of Julia (there is for example a Julia Pkg GitHub issue titled "Allow activating custom project targets" which should improve this).

Do you mean `Pkg.test()`? That works as usual, and I just pushed a fix so that it works both if run from the root project (NonlinearEigenproblems) and from the test project.

maxbennedich:
Below is how a test can be run after this change. First start julia with the `test` project activated, e.g.:
```
julia --project=test
```
Then include the test you want to run. No need to import the `NonlinearEigenproblems` project, since that's done by the test.
```julia
julia> include("test/core.jl")
```
Note that in order to run tests, the `test` project first has to be built so that all the dependencies are installed. By only building the `NonlinearEigenproblems` project, the test dependencies will not be installed (this was the whole point behind this rewrite). To build the tests, one can do something like the below (this is also in `.travis.yml`). Note that the test project depends on the `NonlinearEigenproblems` project, so there's no need to build both projects. There shouldn't be a need to do this more than once, but if installing in a clean environment, such as on `eight`, you probably need to do this. And possibly also if new dependencies are added to either project.
```
julia --project=test -e 'using Pkg; Pkg.build()'
```

PR #125, opened 2018-10-02
Correctly test for warnings in types test

maxbennedich:
The previous test did not test for warnings correctly.

I also renamed the test to `neptypes.jl` to make it a bit more descriptive, and match the testset name.


eringh:
Good renaming!

What was the problem with the warning? I thought that regexp would catch any warning, regardless of message. Is that a bad practice?

maxbennedich:
It was not the message, but the usage of `@test_warn`. In Julia 0.7+, one should use `@test_logs` instead. Search for a github issue named "@test_warn in Julia 0.7 and 1.0" for some details.

I think it's slightly better to specify a message, for two reasons: 1) to ensure that the test does what it's intended to. In this case, the test passed even though it didn't see the warning, since the regex ".*" matched the empty string, and 2) the test becomes more informative (self-documenting code).

(The disadvantage is that if someone rewords the message, the test fails; IMO that's ok, it's easy to spot the failing test and update the message.)

jarlebring:
I think @meleg anyway wanted to change the SPMF-consistency warning  to an error?

meleg:
@jarlebring actually not. I had an error commented just to see the tests that failed (warnings can be hard to see). I am fixing this right now.

eringh:
Are we ready to merge this?

PR #127, opened 2018-10-02
Use verify_lambdas in some tests

maxbennedich:
It provides a convenient and consistent way to test that the expected number of eigenvalues converged, and that the residual norms of all solutions are within the desired tolerance.

Note that this implementation uses `default_errmeasure` to compute the norm, which is defined as `norm(compute_Mlincomb(nep, λ, v)) / norm(v)`. Most tests rather used an absolute norm (not dividing by `norm(v)`). I assumed that this was quite arbitrarily chosen, and that the relative norm is usually a better indicator, but please let me know if this was intentional and I'll update or revert the change in this PR.

### DIFF
--- a/test/broyden.jl
+++ b/test/broyden.jl
@@ -13,8 +13,5 @@ using LinearAlgebra
     # Test by computing the eigenpairs
     λv,X = eigen(S)
     V_eigvecs=V*X;
-    for k=1:size(λv,1)
-        normalize!(V_eigvecs[:,k])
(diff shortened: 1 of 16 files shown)


eringh:
I think this looks like a good change. Increasing code share and with a common way of testing.
Also, relative residual norm seems more reasonable than absolute residual norm to me.

maxbennedich:
Thanks for looking into it. Merging.

Issue #120, opened 2018-09-27
Dependency AbstractAlgebra and PolynomialZeros


jarlebring:
I'm afraid we will have to deactivate or try to get around the use of PolynomialZeros which depends on AbstractAlgebra. The "welcome to" message is not acceptable for us. I don't think they will remove it any time soon considering the responses to my request to remove it on their github (see issue 158 in  https://github.com/Nemocas/AbstractAlgebra.jl). We can look into other polynomial root finder packages.

meleg:
Naive question: can we just use a naive approach like computing the eigenvalues of a (scaled) companion matrix as root-finder? We usually do not have very ill-conditioned polynomial (as far as I know).

jarlebring:
One of the reasons we moved to this particular package was its support for BigFloat. `eigen` doesnt support BigFloat so that naive approach would need some "hacks" to work.

maxbennedich:
We can create a new package called `AbstractAlgebraWithoutTheWelcomeMessage.jl`, which is a copy of theirs but with the welcome message removed. It can be automated to always stay in sync with their repository. Their license seems to allow it. 😛

meleg:
I suggest one of these options:
- We implement our own polynomial root finder (maybe we can simply use the Ehrlich-Aberth method), We should be able to do it in BigFloat.
- We use the eigenvalue of compation matrix (and we do not support BigFloat)
- We try another package

I am slightly more in favor of the first option (and volunteer for eventually trying).

jarlebring:
Approach 2 can be fixed to work for bigfloats. Basically do an eigen implementation for BigFloat matrices: Round to Float64 => use eigen on Float => Use result as starting value for two steps of inverse iteration. 

I would be slightly in favor of approach 2, since the changes seem the smallest and removes a dependency. 

The EA method is not the most common approach for poly roots, and I see a high risk of a hidden amount of work required to design starting values and treat special cases.

jarlebring:
I will close this and maybe revisit it if the abstract algebra package changes behavior.

PR #129, opened 2018-10-03
Naive poly_roots 

jarlebring:
This contains a naive fix for #120 where we rely on companion matrix construction and do a couple of steps of Rayleigh qoutient iteration with output of `eig(::Matrix{Float64})` as starting values . We only used `poly_roots` in one place, so the fix is not so intrusive.

### DIFF
--- a/src/NEPTypes.jl
+++ b/src/NEPTypes.jl
@@ -570,8 +570,14 @@ julia> compute_Mder(pep,3)-(A0+A1*3+A2*9)
         for i=1:size(nep.A,1)
             a[i]=dot(y,nep.A[i]*x);
         end
-        p=Poly(a);
-        rr=real(poly_roots(p));  # Only works if polynomial roots are real
+        m=size(nep.A,1);
+        A=zeros(T,m-1,m-1);
+        for k=1:m-1
+            A[ķ,k+1]=1;
+            A[end,k]=-a[k]/a[end]
+        end
+        pp=eigvals(A);
+        rr=real(pp);  # This function only works if polynomial roots are real
         return rr
     end
 
--- a/src/NEPTypes.jl
+++ b/src/NEPTypes.jl
@@ -578,9 +578,12 @@ julia> compute_Mder(pep,3)-(A0+A1*3+A2*9)
             A[end,k]=-a[k]/a[end]
         end
         pp=eigvals(A);
-        rr=real(pp);  # This function only works if polynomial roots are real
-        II=sortperm(abs.(rr .- target))
-        return rr[II];
+        if (T <: Real) # If we specify real arithmetic, return real eigvals
+            pp=real(pp);
+        end
+
+        II=sortperm(abs.(pp .- target))
+        return pp[II];
     end
 
     # Overload a version of eigvals(::Matrix{BigFloat}) for compute_rf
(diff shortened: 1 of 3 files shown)


src/NEPTypes.jl
```diff
@@ -570,11 +568,47 @@ julia> compute_Mder(pep,3)-(A0+A1*3+A2*9)
...
+        # Build a bigfloat companion matrix
+        A=zeros(T,m-1,m-1);
+        for k=1:m-1
+            if (k<m-1)
+                A[k,k+1]=1;
+            end
+            A[end,k]=-a[k]/a[end]
+        end
+        pp=eigvals(A);
+        rr=real(pp);  # This function only works if polynomial roots are real
```
eringh: Is this needed? I see that we had it before. But why can't the Rayleigh quotient be complex?
Shouldn't we rather make it dependent on the arithmetic type `T`?

jarlebring: I will add some code which does this conditionally based on `T`. I am not sure we have ever carefully defined the meaning of `T` in  `compute_rf`. It is currently not clear to me if it is an arithmetic type (and not related to returned type) or the expected return type or some other type logic in relation to input.

src/NEPTypes.jl
```diff
@@ -570,11 +568,50 @@ julia> compute_Mder(pep,3)-(A0+A1*3+A2*9)
...
+        T=eltype(A);
+        Tfloat= (T <: Complex) ? (ComplexF64) : (Float64)
+        Afloat=Matrix{Tfloat}(A);
+        λv,X=eigen(Afloat);
+        local λv_bigfloat=Vector{Complex{BigFloat}}(λv);
+        # Some Rayleigh quotient iteration steps in bigfloat to improve accuracy:
+        for j=1:size(λv,1)
+            local s=λv_bigfloat[j];
+            z=Vector{Complex{BigFloat}}(X[:,j])
+            for f=1:3
```
maxbennedich: Could the number of iterations be figured out dynamically, by residual norm or something? (Or perhaps there's no need for that?)

meleg:
A small issue. Is this method numerically stable for computing roots of an high degree polynomial (say > 10)? My suggestion is to not use the companion matrix but the scaled companion matrix (iar-style). In this way the eigenvectors will not have large elements. I remember from a matlab testing that the scaled companion often gives better results (or at least never worst). Maybe it is not necessary, if you tested and it works, it is fine.

src/NEPTypes.jl
jarlebring: In principle yes, but they are always heuristic. The decision to take three is however already from theory. The Rayleigh quotient iteration has cubic convergence. If we initiate with an approximation with error 10^-16:

Error after one step (10^-16)^3=10^-49 
Error after two steps (10^-49)^3=10^-145 which is smaller than eps(BigFloat)=10^-77. 

I take one extra step to be "sure".

jarlebring:
I think if you take a different companion linearization we will get stability problems elsewhere. For IAR there is a more natural choice since we are interested in eigenvalues close to the origin. If we start looking at stability there is an enormous depth of research about how to do companion linearization, and specialized algorithms. Very difficult to do "correctly". I therefore prefer to do a naive approach (w standard companion matrix), rather than trying to incorporate research in an incomplete way.

maxbennedich:
Nice job! Should we add a unit test for this, or is it covered by existing tests?

jarlebring:
This is a drop-in replacement. I think it is already quite well covered.

jarlebring:
I will merge this as preparation for the release. Maybe we reactivate PolynomialZeros in the next release.

Issue #139, opened 2018-10-12
Type vs DataType


eringh:
We have a few occurrences of `DataType` in the package. I think, after getting into some troubles with that and reading this [post](https://stackoverflow.com/questions/42283820/julia-diffrence-between-type-and-datatype), that we might want to change it to `Type`.

Anyone that has a different understanding?

Issue #95, opened 2018-09-19
Make LinSolvers type stable 


jarlebring:
At the meeting 2018-09-19, we said we wanted to make LinSolvers type stable. I have not been able to generate an example where it DefaultLinSolver or BackslashLinSolver fail `@code_warntype`.

jarlebring:
Ping @eringh

eringh:
That is great!
Maybe I was a bit unclear in my thoughts, sorry for that. I would argue that is as it should, since both of them are completely specified by their respective parametric types, and contain no un-typed or abstractly typed fields (as described [here](https://docs.julialang.org/en/latest/manual/performance-tips/#Avoid-fields-with-abstract-type-1).

However, I think that `BackslashLinSolver` can be reformulated as:
```
    struct BackslashLinSolver{T_mat<:AbstractMatrix} <: LinSolver
        A::T_mat
    end
```
i.e., there is no need for having the `T_num<:Number` parameterizing it (if my understanding for the language is better compared to when I introduced it). The `DefaultLinSolver` got a tidy-up in #84, which I think was great.

Perhaps we need to look more closely on the `GMRESLinSolver`. I can take that at a later stage. The `EigSolver` should also get a similar fix I think.

jarlebring:
Okay. Sounds like lower priority. I expect the efficiency gain for methods using EigSolver to be quite small since it is so expensive computationally to solve eigenvalue problems.

eringh:
f9e193a41de90ff7da9113db19c74d88902f519f

eringh:
Why are some linsolvers/eigsolvers `struct` and some `mutable struct`?
I don't think we will update the parameters of the solvers after creation, or will we?

eringh:
I do not manage to get type-stability for the `NativeEigSSolver`. However, the code 
```
        if(nev == 1)
            return D[1],V[:,1]
        end
```
breaks the type-stability for the `NativeEigSolver`. Do we want to keep it anyway?

maxbennedich:
> Why are some linsolvers/eigsolvers `struct` and some `mutable struct`?
> I don't think we will update the parameters of the solvers after creation, or will we?

I generally prefer immutability (`struct`). Even if parameters are changed after creation, an option would be to create a new instance. Of course there are exceptions, but I think it's good practice to make types immutable by default.

maxbennedich:
[...]

PR #142, opened 2018-10-12
Lineigensolver and DataType

eringh:
Did a re-work of the eig-solvers, and some of the lin-solvers; see #95. Also worked on the `DataType` vs `Type`; see #139.

There are some questions remaining in #95, but this is approaching ready. Although I did not really manage to get type-stability, I believe the design is still favorable compared to the old one. At least for `NativeEigSolver` we know where the type-stability breaks, and for `NativeEigSSolver` I get that `eigs()` does not seem to be type stable (and hence I see no hope for us).


test/linsolver.jl
```diff
@@ -90,4 +90,74 @@ using LinearAlgebra
...
+
+        for i = 1:3
+            @test λ1[i] ≈ λ2[i]
+        end
+
+        @info "NativeEigSSolver"
+        Random.seed!(0)
+        n = 20
+        A = sprand(ComplexF64, n, n, 0.25) + sparse(1.0*I,n,n)
+        B = sprand(ComplexF64, n, n, 0.25) + sparse(1.0*I,n,n)
```
maxbennedich: You can change `sparse(1.0*I,n,n)` to just `I`.

eringh: Good point! I changed it.
(I also did separate tests of all parts of increase encapsulation of the tests)

eringh:
Now I start to feel ready with this work. I did not really manage to get type stability, but I have some problems with the type stability of the underlying functions. Perhaps I am calling them in an inappropriate way, or they are just unstable. It seems rather stable, except in terms of `Float64`/`ComplexF64`,for the generalized and dense problem.

maxbennedich:
Looks like a nice cleanup to me!

jarlebring:
:+1:

PR #146, opened 2018-10-23
Der spmf

meleg:
A new `nep` subtype is introduced: `DerSPMF` as subtype of `AbstractSPMF`.

The idea behind this type is the following: in several methods it is needed to call many times the function `compute_Mlincomb` with the same shift. What it happens now is that, for each call of this function, all the derivatives of the nonlinear functions are "implicitly" computed.

A way to use less memory and speed up `compute_Mlincomb` (under the assumption that we call this function always with the same shift) is to pre-compute the derivatives of the nonlinear functions. There are at the moment two ways to construct a `DerSPMF`.

One way is to simply call `Dnep=DerSPMF(nep,σ,m)` where `nep` is a `SPMF_NEP` and `m` is the number of wanted derivatives. In this case the derivatives will be precomputed with matrix function evaluations. Since the matrix function evaluation method cannot always be used (and sometimes is unstable), the user can provide the derivatives `fD` (matrix) as input  `Dnep=DerSPMF(nep,fD,σ)`. 

Main files are:
- profiling/DerSPMF_test.jl
- profiling/ilan_with_DerSPMF.jl

A quick test shows that the new `compute_Mlincomb` is about 3 times faster (see the following comparisons: old vs new `compute_Mlincomb`). The script to generate these results is profiling/DerSPMF_test.jl
```
  378.628 ms (455 allocations: 412.52 MiB)
  104.318 ms (25 allocations: 68.67 MiB)
```

If use this function inside a method we also get a substantial speed up. In order: ilan (new approach), ilan (old) and tiar (old). The script to generate these results is profiling/ilan_timed.jl.
```
  8.187049 seconds (1.73 M allocations: 11.287 GiB, 25.28% gc time)
 12.806317 seconds (4.00 M allocations: 13.245 GiB, 22.47% gc time)
 15.375965 seconds (3.69 M allocations: 27.696 GiB, 13.53% gc time)
```

Can I further improve `DerSPMF` and `compute_Mlincomb`? 
All feedback are welcome.


profiling/DerSPMF_test.jl
```diff
@@ -0,0 +1,70 @@
+using NonlinearEigenproblems, Random, SparseArrays, Revise, LinearAlgebra, BenchmarkTools
+import ..NEPTypes.AbstractSPMF
+struct DerSPMF{T<:AbstractMatrix,FDtype} <: AbstractSPMF{T}
+    spmf::SPMF_NEP{T}
+    fD::Matrix{FDtype}
+    σ::Number
```
jarlebring: `Number` is not a concrete type so you will not get type stability. You probably need it the type parameters (but not in the constructor).

profiling/DerSPMF_test.jl
```diff
@@ -0,0 +1,70 @@
...
+end
+
+## one constructor takes spmf as input and compute the derivatives
+function DerSPMF(spmf::SPMF_NEP,σ::Number,m::Int)
+      # Compute DD-matrix from get_fv(spmf)
+      TT=promote_type(typeof(σ),eltype(nep.A[1]))
+      p=length(nep.fi)
+      fD=zeros(TT,m,p)
+      # matrix for the computation of derivatives
+      SS=diagm(0 => σ*ones(TT,2m+2)) + diagm(-1 => (1:2m+1))
```
jarlebring: Addition of matrices can be avoided. Why not?  
```
diagm(0=> σ*ones(TT,2m+2),  -1 => (1:2m+1))
```

profiling/DerSPMF_test.jl
```diff
@@ -0,0 +1,70 @@
...
+    p=size(nep.fD,2)
+    n,k=size(V)
+    # Type logic
+    TT=promote_type(eltype(V),typeof(λ),eltype(nep.spmf.A[1]),eltype(a))
+    z=zeros(TT,n)
+    afD=a.*nep.fD[1:k,:];
+    @inbounds for j=1:p
+        z .+= nep.spmf.A[j]*(V*afD[:,j])
+    end
+    return reshape(z,n)
```
jarlebring: reshape should not be necessary?

profiling/DerSPMF_test.jl
```diff
@@ -0,0 +1,70 @@
...
+    local n,k,p
+    p=size(nep.fD,2)
+    n,k=size(V)
+    # Type logic
+    TT=promote_type(eltype(V),typeof(λ),eltype(nep.spmf.A[1]),eltype(a))
+    z=zeros(TT,n)
+    afD=a.*nep.fD[1:k,:];
+    @inbounds for j=1:p
+        z .+= nep.spmf.A[j]*(V*afD[:,j])
+    end
```
jarlebring: Would it be faster to also precompute `VafD=V*afD` before the for-loop?  Matrix matrix products are in general faster.

jarlebring:
At the moment this code essentially only changes things in the profiling directory not the src directory. Was this intentional? I would say it is better if you already try to make it a part of the src.

maxbennedich:
Nice job! Below are some general ideas to increase performance. Note that these don't always have to be applied -- as always, I'd recommend focusing on writing clear code, and then use profiling to determine where there are performance problems, and optimize those portions only.

- Use `view` when referencing a part of a matrix to prevent creating a copy
- Use `Matrix{T}(undef, n, m)` instead of `zeros(T, n, m)` to allocate matrices when you don't need to zero all elements
- For indexing that will be executed a large number of times (such as devectorized for loops), consider `@inbounds`

profiling/DerSPMF_test.jl
```diff
@@ -0,0 +1,70 @@
...
+
+## one constructor takes spmf as input and compute the derivatives
+function DerSPMF(spmf::SPMF_NEP,σ::Number,m::Int)
+      # Compute DD-matrix from get_fv(spmf)
+      TT=promote_type(typeof(σ),eltype(nep.A[1]))
+      p=length(nep.fi)
+      fD=zeros(TT,m,p)
+      # matrix for the computation of derivatives
+      SS=diagm(0 => σ*ones(TT,2m+2)) + diagm(-1 => (1:2m+1))
+      fD=zeros(2*m+2,p)
```
maxbennedich: I think this should have a type, like on line 14. And line 14 could be removed, right?

meleg:
@jarlebring and @maxbennedich I have changed the code according to your suggestions. Thanks a lot, I got very good improvements in memory and time. Compare the following with the first post.

The result of profiling/DerSPMF_test.jl which compare `compute_Mlincomb` for the type `SPMF_NEP`  and the new specialized version for `DerSPMF` is
```
  378.628 ms (455 allocations: 412.52 MiB)
  104.318 ms (25 allocations: 68.67 MiB)
```
If we use this function inside a method we also get a substantial speed up. In order: ilan (new approach), ilan (old) and tiar (old). The script to generate these results is profiling/ilan_timed.jl.
```
  8.187049 seconds (1.73 M allocations: 11.287 GiB, 25.28% gc time)
 12.806317 seconds (4.00 M allocations: 13.245 GiB, 22.47% gc time)
 15.375965 seconds (3.69 M allocations: 27.696 GiB, 13.53% gc time)
```

Feel free to continue giving feedback. I will try to optimize a bit more and comment the new code.

src/NEPTypes.jl
```diff
@@ -1327,4 +1328,59 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
+    end
+    function get_fv(nep::DerSPMF)
+        return get_fv(nep.spmf)
+    end
+    function get_Av(nep::DerSPMF)
+        return get_Av(nep.spmf)
+    end
+
+    ## one constructor takes spmf as input and compute the derivatives
+    function DerSPMF(spmf::SPMF_NEP,σ::Number,m::Int)
```
jarlebring: Can't `SPMF_NEP` be replaced by `AbstractSPMF` and use the corresponding get-functions?

jarlebring:
The change to have the constructor take an `AbstractSPMF` makes a potentially useful feature possible: One can use it on itself, with different shift, such that one can provide precomputed values for several points. This is actually needed in rational Krylov methods (when only a few shifts are used). 

```
nep=SPMF_NEP(...)
Dnep=DerSPMF(nep,σ1,100) 
Dnep2=DerSPMF(Dnep,σ2,100)
```
After this `Dnep2` should use precomputed values for evaluations in either σ1 and σ2.

jarlebring:
By the way: The experiments look promising!

meleg:
@jarlebring  I have fixed the code according to the last two comments.

jarlebring:
Looks good. I think only one documentation describing both constructors is clearer, otherwise ?DerSPM becomes repetitive. This is also recommended, see third sentence in https://docs.julialang.org/en/v1/manual/documentation/#Functions-and-Methods-1  "In general, only the most generic method should be documented" Please also add some unit tests.

src/NEPTypes.jl
```diff
@@ -1327,4 +1328,96 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
+          # Compute DD-matrix from get_fv(spmf)
+          Av=get_Av(spmf)
+          fv=get_fv(spmf)
+          TT=promote_type(typeof(σ),eltype(Av[1]))
+          p=length(fv)
+          # matrix for the computation of derivatives
+          SS=diagm(0=> σ*ones(TT,2m+2),  -1 => (1:2m+1))
+          fD=Matrix{TT}(undef, 2*m+2,p)
+          for t=1:p fD[:,t]=fv[t](SS)[:,1] end
+          return DerSPMF(SPMF_NEP(Av,fv),σ,fD);
```
jarlebring: Why do you recreate the SPMF_NEP? I think it is better to make DerSPMF parametric with respect to the SPMF_NEP type and do
```
           return DerSPMF(spmf,σ,fD); 
```
The feature I described will not work with the current implementation since you convert it to SPMF_NEP.

meleg: @jarlebring I have changed (in the last commits) the code in a way to not recreate a `SPMF_NEP`. I did not do that in the way you suggested, since I realized that I wanted the `DerSPMF` creator to take as input any `AbstractSPMF` (and not only `SPMF_NEP`). For example I would like to take a `DEP` as input. 

With this change I cover your suggestion (I think).

src/NEPTypes.jl
```diff
@@ -1327,4 +1328,97 @@ Returns true/false if the NEP is sparse (if compute_Mder() returns sparse)
...
+                        V::AbstractVecOrMat,
+                        a::Vector=ones(size(V,2))) where {T,FDtype}
+
+        if λ!=nep.σ
+            return compute_Mlincomb(nep.spmf,λ,V,a)
+        else
+
+            local n,k,p
+            p=size(nep.fD,2)
+            n,k=size(V)
```
jarlebring: This will not work if V is a vector (and it can be a VecOrMat), should probably be
```
n=size(V,1); k=size(V,2);
```

src/NEPTypes.jl
jarlebring: Looks good. Should be easy to test based on your example:
```
A0=[1 3; 4 5]; A1=[3 4; 5 6];
id_op=S -> one(S) # Note: We use one(S) to be valid both for matrices and scalars
exp_op=S -> exp(S)
nep=SPMF_NEP([A0,A1],[id_op,exp_op]);
m=5
σ1=3.5;
Dnep=DerSPMF(nep,σ1,m)
σ2=9;
DDnep=DerSPMF(Dnep,σ2,m)
V=randn(2,2);
compute_Mlincom(nep,0,V)-compute_Mlincom(Dnep,0,V)
compute_Mlincom(nep,σ1,V)-compute_Mlincom(Dnep,σ1,V)
compute_Mlincom(nep,σ2,V)-compute_Mlincom(Dnep,σ2,V)
compute_Mlincom(nep,0,V)-compute_Mlincom(DDnep,0,V)
compute_Mlincom(nep,σ1,V)-compute_Mlincom(DDnep,σ1,V)
compute_Mlincom(nep,σ2,V)-compute_Mlincom(DDnep,σ2,V)
```

meleg:
I changed the code according to the last suggestions. I have initiated a test for the new code. I will not comment it now, I will do it in a later stage. If there are no more errors I would like to merge it.

jarlebring:
:+1:

PR #148, opened 2018-12-05
BEM Fichera corner hardcoded

jarlebring:
This is a reimplementation of the boundary element method applied to Helmholtz equation on a cubic domain except one corner removed (Fichera corner). It is based on the model and some code in https://link.springer.com/article/10.1007/s10543-012-0381-5 where they use a Chebyshev interpolation approach. Only the BEM-problem not the Chebyshev interpolation method is implemented. The implementation is for `compute_Mder` (any derivative) and the other compute functions are available with the `compute_Mlincomb_from_Mder`.

This is a dense and very computationally demanding NEP. 

```julia
julia> nep=nep_gallery("bem_fichera",5);
julia> size(nep)
(864, 864)
julia> @btime compute_Mder(nep,6.0);
  10.269 s (66149148 allocations: 8.77 GiB)
```
Several methods have been tested. Among the newton-type methods, `mslp` seems to be the most reliable. IAR works but is slow. 
```julia
julia> (λ,v)=mslp(nep,λ=8,displaylevel=1)
Iteration:1 errmeasure:4.122635537095636e-6 λ=8.128272919317748 + 0.007584851218214716im
Iteration:2 errmeasure:1.787963303973586e-8 λ=8.132181234599427 - 1.952792817964521e-5im
Iteration:3 errmeasure:3.2884958163572594e-13 λ=8.132145310156643 - 1.2648247028455485e-5im
Iteration:4 errmeasure:4.6607986030841e-18 λ=8.132145310195453 - 1.264891804832194e-5im
(8.132145310195453 - 1.264891804832194e-5im, Complex{Float64}[3.08473e-5-9.8713e-6im, 9.46458e-5+2.08586e-5im, -0.000418303-9.3624e-5im, -2.27161e-5+3.2045e-5im, -0.00168228-0.000446522im, -0.00660488-0.0018462im, -0.00705554-0.00195021im, -0.000714245-0.000123651im, -0.010653-0.00296256im, -0.0250155-0.00702815im  …  0.00369925+0.00101557im, 0.025547+0.00717101im, 0.0333126+0.00931856im, 0.0158614+0.00438927im, 0.00325204+0.000835354im, 0.021329+0.00595943im, 0.0126512+0.0034611im, 0.00130882+0.000172086im, 0.00131286+0.000207463im, 0.0125435+0.00344975im])
```

The discretization is very hard-coded, but works well for a benchmark. Rather than providing more advanced discretizations/domains, we may later consider linking with the [BEM++ package](https://bempp.com/).

### DIFF
--- a/src/gallery_extra/bem_hardcoded/assemble_BEM.jl
+++ b/src/gallery_extra/bem_hardcoded/assemble_BEM.jl
@@ -1,0 +1,84 @@
+include("deHoop.jl");
+function assemble_BEM(lambda, mesh, gauss_order,der=0)
+    # define Gauss-Legendre quadrature points and weights
+    local pt,wg;
+    if (gauss_order==3)
+	# points
+	pt = [2/3 1/6 1/6;
+	      1/6 2/3 1/6;
+	      1/6 1/6 2/3];
+	# weights
+        wg = [1/3;
+              1/3;
+              1/3];
+    else
+        error("The Gauss quadrature order you specified is not implemented")
+    end
+
+    # Number of triangles = size of matrix
+    n = size(mesh,1);
+
+    # Storage for the computed matrix
+    T = zeros(ComplexF64,n, n);
+
+    # prepare auxiliary vectors
+    rowind = vec(repeat(1:gauss_order,1,gauss_order)'); # [1 1 1 2 2 2 3 3 3 ]
+    colind = vec(repeat(1:gauss_order,1,gauss_order))  # [1 2 3 1 2 3 1 2 3]
+
+    # pre-compute guassP and guassW for all triangles
+    for tri = mesh
+	VK = [tri.P1'; tri.P2'; tri.P3'];
+	tri.aux.gaussP = transpose(pt * VK);
+	tri.aux.gaussW = tri.area * wg;
+    end
+
+    for row = 1:n
+	@inbounds for col = row:n # Skip because of symmetry
+
+	    # compute (row, col)-entry of T
+            trir=mesh[row];
+            tric=mesh[col]
+
+
+            # Compute a distance vector where zero elements
+            # are set to one and marked in the idx vector
+            S1=trir.aux.gaussP[:, rowind];
+            S2=tric.aux.gaussP[:, colind];
+            s0=(S1 - S2).^2
+	    dist = sqrt.(sum(s0, dims=1));
+	    idx = (dist .== 0);
+	    dist[idx] .= 1;
+
+
+            # E = derivative of greens function
+            local E
+            if (der==0) # Zero'th derivative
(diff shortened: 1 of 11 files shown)


jarlebring:
This how you can visualize the solution:
```julia
using NonlinearEigenproblems, PyPlot
nep=nep_gallery("bem_fichera")
(λ,v)=mslp(nep,λ=8.1,displaylevel=1)
v=v./maximum(abs.(v));
for k=1:size(nep.mesh,1);
    tri=nep.mesh[k];
    x=[tri.P1[1];tri.P2[1];tri.P3[1];tri.P1[1]];
    y=[tri.P1[2];tri.P2[2];tri.P3[2];tri.P1[2]];
    z=[tri.P1[3];tri.P2[3];tri.P3[3];tri.P1[3]];
    col=[1-abs.(v)[k];0;0]; # plot abslolute value
    X=[x[1] x[2]; x[3] x[3]];
    Y=[y[1] y[2]; y[3] y[3]];
    Z=[z[1] z[2]; z[3] z[3]];
    plot_surface(X,Y,Z,color=col,alpha=0.8);
    plot_wireframe(X,Y,Z,color=[0;0;0],linewidth=1,alpha=0.5,);
end
```
This is the absolute value:
fichera

src/gallery_extra/bem_hardcoded/assemble_BEM.jl
```diff
@@ -0,0 +1,84 @@
+include("deHoop.jl");
+function assemble_BEM(lambda, mesh, gauss_order,der=0)
```
maxbennedich: Could the pre-computed code below that's not dependent on the lambda be stored / cached in `BEM_NEP`? I'm thinking so that they wouldn't have to be re-calculated each time `compute_Mder` is run. (But perhaps this is not a bottleneck anyway.)

src/gallery_extra/bem_hardcoded/genmesh.jl
```diff
@@ -0,0 +1,113 @@
+# adds a triangle to mesh where a b c d are the relative index coordinates
+macro addtri(a,b,c,d)
```
maxbennedich: Nice macro!

src/gallery_extra/bem_hardcoded/assemble_BEM.jl
jarlebring: ThanksA It's not a bottleneck, but I did it anyway since it seems cleaner to also have precomputation for the assembly (not just the mesh generation).

jarlebring:
The author of the code I used as inspiration is apparently, M. Steinlechner and his thesis http://sma.epfl.ch/~anchpcommon/students/steinlechner.pdf. References have been updated.

jarlebring:
Code mature. Merging.

PR #149, opened 2018-12-07
movebc tutorial

jarlebring:
A tutorial containing how to construct NEPs by moving boundary conditions. Visualization of the result.

newplot

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -13,7 +13,8 @@ makedocs(
         "LinSolver" => "linsolvers.md",
         "NEP Transformations" => "transformations.md",
         "NEP Gallery" => "gallery.md",
-        "Tutorial 1 (BEM)" => "bemtutorial.md"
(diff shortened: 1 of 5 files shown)


jarlebring:
One more figure needed:
newplot2

maxbennedich:
Cool! This type of tutorial seems very useful.

May I ask out of curiosity where/how you stored the images? In the past I've usually made markdown images part of the repository (if they don't take up much space), to make the repo self-contained, but I'm a bit torn over if this is a clean way to do it or not. (Besides, larger images/videos/resources are likely better kept outside the repo to keep its size down.)

jarlebring:
Yes... I decided if we want people to use it, we need to elaborate on applications we can solve.

I took inspiration from oth julia packages using Documenter.  A common technique seems to be to post the image here (in a PR or issue) and then copy-paste the link to the image.  I don't know if the images can ever be removed or replaced, so this also does not seem very clean.

PR #152, opened 2019-01-06
Deflation tutorial

jarlebring:
A new tutorial illustrating how to do deflation in NEP-PACK. It is already visible in the online documentation: https://nep-pack.github.io/NonlinearEigenproblems.jl/deflate_tutorial/


eringh:
I think it looks very nice!

PR #160, opened 2019-02-10
Simple NEP creator helpers 

### LINKED ISSUE #159: Convenience functions/types for reduced compute functions types
As you can see in the tutorials https://nep-pack.github.io/NonlinearEigenproblems.jl/tutorial_matlab1/ and https://nep-pack.github.io/NonlinearEigenproblems.jl/tutorial_call_python/ one currently has to define new NEP-types if one has an external compute_Mder function (or compute_Mlincomb). This seems a bit complicated considering how common it is. 

We could add a `compute_Mder_NEP` which takes a function as input, such that the MATLAB tutorial example would be:
```
julia> nep=compute_Mder_NEP((s,v) -> mat"compute_derivative_k(double($s),double($der))",10)
```
It would then default to using `compute_Mlincomb_from_Mder`, maybe also something similar for `compute_MM`.

jarlebring:
This should resolve #159. It adds two creators:
```
Mder_NEP
Mder_Mlincomb_NEP
```
Simple example from documentation:
```julia
julia> using LinearAlgebra; # For the I function
julia> function my_Mder(s,der)
    A0=ones(Float64,3,3)-I; A0[1,1]=-1;
    A1=ones(Float64,3,3)*3; A1[2,3]=0;
    if (der==0)
       return A0+A1*s;
    elseif (der==1)
       return A1;
    else
       return zero(A0);
    end
end
julia> nep=Mder_NEP(3,my_Mder);
julia> (λ,v)=augnewton(nep,v=ones(3));
julia> norm(compute_Mder(nep,λ)*v)
5.551115123125783e-17

```
The tutorials have only been updated with a note that this can be done easier in the development version. We can update the tutorials after the next release.

### DIFF
--- a/docs/src/bemtutorial.md
+++ b/docs/src/bemtutorial.md
@@ -113,3 +113,7 @@ for k=1:size(nep.mesh,1);
     plot_wireframe(X,Y,Z,color=[0;0;0],linewidth=1,alpha=0.5,);
 end
 ```
+
+Note: The above functionality can also be achieved with  `Mder_NEP` in the development version of NEP-PACK
+
+![To the top](http://jarlebring.se/onepixel.png?NEPPACKDOC_BEMTUTORIAL)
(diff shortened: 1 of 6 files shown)


maxbennedich:
Nice! Seems useful.

PR #168, opened 2019-03-08
errmeasure types

### LINKED ISSUE #150: Passing `errmeasure` as type
As proposed in the Issue  #39, it would be convenient to pass `errmeasure` as type and not as function. We can follow the instructions that @jarlebring gave in a comment.

> It would be nice to have if we can have several easily accessible ways to measure the error:
> `DefaultErrmeasure`
> `ResnormErrmeasure`
> `BackwardErrmeasure` (for spmfs:  Resnorm / (||A_1|| |f_1(λ)|+... ||A_m|| |f_m(λ)|)
> `ReferenceErrmeasure` measure distance to a precomputed reference solution.  
> 
> The interface to compute an errmeasure could be  e.g.
> ```
> function eval_errmeasure(::Type{ResnormErrmeasure},nep,λ,v)
>    return norm(compute_Mlincomb(nep,λ,v))/norm(v);
> end
> ```
> [...]
> I think `errmeasure` is certainly the easiest one, but still not trivial. It would be nice to have if we can have several easily accessible ways to measure the error:
> 
> That would be enough for `ResnormErrmeasure`, but we may need more functions.  Note that `BackwardErrmeasure` would ideally compute ||A_1|| ... || A_m|| before the main iteration loop. I don't see how that could be done without also adding a `precompute_data` function (like we now have for `y0`-computation in `iar_chebyshev`). `ReferenceErrmeasure` would need the user to provide the reference solution information somehow. 
> [...]
> 

_Originally posted by @jarlebring in https://github.com/nep-pack/NonlinearEigenproblems.jl/issues/39#issuecomment-404179942_

My suggestion is to use the name `RelativeErrmeasure` instead of `BackwardErrmeasure`.

In this way the user can define is own error measurement by defining a new type and by overloading the function  `errmeasure` by following the same logic of the orthogonalization (`orthmethod`), for example, in `iar`. See the following quoted code from NonlinearEigenproblems.jl/test/iar.jl
      

```
# The user can create his own orthogonalization function to use in IAR
function doubleGS_function!(VV, vv, h)
    h[:]=VV'*vv; vv[:]=vv-VV*h; g=VV'*vv; vv[:]=vv-VV*g;
    h[:] = h[:]+g[:]; β=norm(vv); vv[:]=vv/β; return β
end
# Then it is needed to create a type to access to this function
abstract type DoubleGS <: IterativeSolvers.OrthogonalizationMethod end
# And then introduce a function dispatch for this new type in order to use
# the defined orthogonalization function
import IterativeSolvers.orthogonalize_and_normalize!
function orthogonalize_and_normalize!(V,v,h,::Type{DoubleGS})
doubleGS_function!(V, v, h) end
```

jarlebring:
These are the first steps to resolve #150.

New types are introduces
```
ResidualErrmeasure
BackwardErrmeasure
DefaultErrmeasure  # Becomes BackwardErrmeasure if NEP isa AbstractSPMF otherwise ResidualErrmeasure
```
In order to use errmeasures in a NEP-solver you need to initiate it with a nep first:
```
nep=nep_gallery("nlevp_native_gun");
ermdata=init_errmeasure(ResidualErrmeasure,nep)
```
The `ermdata` contains precomputed errmeasure information (like \|A_1\| ... \|A_m\| ) and to get the error (estimate) do
```
estimate_error(ermdata,s,v)
```
I have only changed the iar-method so far. I want to make sure we get it right before changing all the others.

It seems to "work" since the default behavior solves problems now which were not solved before:
```julia
julia> (λ,v)=iar(nep,σ=200^2, γ=50^2,tol=1e-8);

julia> λ
6-element Array{Complex{Float64},1}:
  44259.41857472756 + 3.575987047379382im 
 43857.600899177254 + 20.52553168228922im 
  48788.73196043914 + 6.323957778415443im 
 48142.068793449165 + 41.89143413922458im 
  22345.04691041269 + 0.6340314337432504im
  54550.39890150822 + 459.46092126972843im

julia> (λ,v)=iar(nep,σ=200^2, γ=50^2,tol=1e-8, errmeasure=BackwardErrmeasure);

julia> λ
6-element Array{Complex{Float64},1}:
  44259.41857472755 + 3.5759870473780153im
  43857.60089917725 + 20.525531682289184im
  48788.73195992363 + 6.323958019966809im 
  48142.06879500499 + 41.891442316226495im
  54550.59086913872 + 459.4528810229781im 
 22345.020210194154 + 0.6368829115391262im
# This was the default behaviour before
julia> (λ,v)=iar(nep,σ=200^2, γ=50^2,tol=1e-8, errmeasure=ResidualErrmeasure); 
ERROR: No convergence: 'Number of iterations exceeded. maxit=30. Check that σ is not an eigenvalue.' eigenvalue approx:Complex{Float64}[43857.6+20.5255im, 44259.4+3.57599im, 48142.1+41.8914im, 48788.7+6.32396im, 54550.2+459.457im, 22345.1+0.642966im], errmeasure:[2.96099e-7, 3.84396e-7, 2.53831e-5, 4.92474e-5, 0.00250231, 0.00551035]
```

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -11,6 +11,7 @@ makedocs(
         "NEP Methods" => "methods.md",
         "NEP Types" => "types.md",
         "LinSolver" => "linsolvers.md",
+        "Error measure" => "errmeasure.md",
         "NEP Transformations" => "transformations.md",
         "NEP Gallery" => "gallery.md",
         "Tutorial 1 (ABC)" => "movebc_tutorial.md",
--- a/docs/src/errmeasure.md
+++ b/docs/src/errmeasure.md
@@ -1,0 +1,147 @@
+# Measuring the error
+
+**This only works on NEP-PACK version 0.2.6 and onwards**
+
+All iterative algorithms need some form of termination
+criteria. In NEP-PACK, all NEP-solvers provide
+the possibility to specify the desired tolerance,
+as well as how the error is measured or estimated.
+The tolerance is specified in the kwarg  `tol` (which is a real number)
+and the way to measure the error is given in `errmeasure`.
+
+## Standard usage
+
+NEP-PACK comes with several ways to measure errors for many NEP-types.
+
+* `errmeasure=ResidualErrmeasure`: The error is estimated by the use of the residual norm:
+```math
+\mathrm{err}=\frac{\|M(λ)v\|}{\|v\|}.
+```
+* `errmeasure=BackwardErrmeasure`: The error is estimated by using the backward error bounds. This error measure will not work for all NEPs. An implementation is provided for any `AbstractSPMF`. If your NEP is an `AbstractSPMF` with terms:
+  ```math
+  M(λ)=A_1f_1(λ)+\cdots+A_mf_m(λ)
+  ```
+  the error will be estimated by
+  ```math
+  \mathrm{err}=\frac{\|M(λ)v\|}{\|v\|}\frac{1}{\|A_1\|_F|f_1(λ)|+\cdots+\|A_m\|_F|f_m(λ)|}.
+  ```
+  In other words, the `BackwardErrmeasure` is a weighting of the `ResidualErrmeasure`.
+* `errmeasure=DefaultErrmeasure`: When this `errmeasure` is specified, NEP-PACK tries to determine a error measure for you. In general, `BackwardErrmeasure` will be preferred if possible. This behavior may change in future versions of NEP-PACK.
+
+Example: Most NEP-solvers take the `errmeasure` as an kwarg.
+```julia-repl
+julia> nep=nep_gallery("qdep0");
+julia> # Solve the problem to residual norm 1e-8
+julia> (λ,v)=mslp(nep,errmeasure=ResidualErrmeasure,tol=1e-8)
+julia> norm(compute_Mlincomb(nep,λ,v))/norm(v) # It's smaller than tol?
+3.503700819937386e-9
+julia> nep isa AbstractSPMF # Is it an AbstractSPMF so we can use BackwardErrmeasure?
+true
+julia> (λ,v)=mslp(nep,errmeasure=BackwardErrmeasure,tol=1e-10)
+julia> factor=abs(fv[1](λ))*norm(Av[1])+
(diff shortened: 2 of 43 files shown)


jarlebring:
It is also possible to use "reference solutions", although it unfortunately requires an additional user-defined types:

```julia
julia> nep=nep_gallery("dep0");
julia> (λref,Vref)=iar(nep);
julia> struct MyErrmeasure <: Errmeasure; nep::NEP; end
julia> function estimate_error(e::MyErrmeasure,λ,v);
       return minimum(abs.(λref .- λ))/abs(λ);
       end
julia> quasinewton(nep,errmeasure=MyErrmeasure,displaylevel=1,v=ones(size(nep,1)));
Precomputing linsolver
Iteration:  1 errmeasure:Inf, λ=0.0 + 0.0im
Iteration:  2 errmeasure:4.069602702630671365e-01, λ=0.5932899278412939 + 0.0im
Iteration:  3 errmeasure:2.964266029876431729e-01, λ=1.1864225690802832 + 0.0im
...
Iteration: 68 errmeasure:3.048542757039950370e-14, λ=-0.3587189459686377 + 0.0im
Iteration: 69 errmeasure:1.609384369674978238e-14, λ=-0.358718945968621 + 0.0im
```
The quasinewton call uses the relative distance to the closest (precomputed) reference eigenvalue.

jarlebring:
I think I found a way to make the user-defined error measures easier. 
The most recent commit makes it possible to use kwarg errmeasure to be a `<:Type{Errmeasure}` or a `Function`. This means the new syntax extends the old syntax from a user-perspective.

```julia
julia> nep=nep_gallery("dep0");
julia> (λref,Vref)=iar(nep);
julia> quasinewton(nep,errmeasure=(λ,v) -> minimum(abs.(λref .- λ))/abs(λ),
displaylevel=1,v=ones(size(nep,1)));
```

jarlebring:
Okay. This PR is starting to feel ready. It turned out to be possible to make the explanation in the documentation concise, which is a positive indicator of the design: https://nep-pack.github.io/NonlinearEigenproblems.jl/errmeasure/

PR #167, opened 2019-03-07
Deflation rewrite

jarlebring:
This fixes #166 and

* Makes deflation more user friendly

* Makes deflation more efficient, in paricular for sparse matrices.

* Makes deflation of a deflated problem not a recursive deflated problem, but just a modified.

* Introduces three deflation modes: Generic, SPMF, MM. SPMF can be used if the SPMF is short and gives a AbstractSPMF. It has some drawbacks in terms of robustness. Generic seems to be the most efficient in many cases. Fits naturally with Juliaesque type-thinking. 

* Updates the deflation tutorial 

**Simulation 1:**
```julia
nep=nep_gallery("dep0_sparse",20);
n=size(nep,1);
local λ,v;
(λ,v)=quasinewton(nep,v=ones(n),λ=3,tol=1e-11,displaylevel=1,maxit=200)
```
```julia
dnep1=deflate_eigpair(nep,λ,v,mode=:Generic);
@btime quasinewton(dnep1,v=ones(size(dnep1,1)),λ=1+1im,tol=1e-11)
```
  2.638 ms (16661 allocations: 3.28 MiB)
```julia
dnep2=deflate_eigpair(nep,λ,v,mode=:SPMF);
@btime quasinewton(dnep2,v=ones(size(dnep1,1)),λ=1+1im,tol=1e-11)
```
  10.774 ms (54923 allocations: 5.88 MiB)

Old approach:
```julia
dnep3=deflate_eigpair(nep,λ,v,mode=:MM);
@btime quasinewton(dnep3,v=ones(size(dnep1,1)),λ=1+1im,tol=1e-11)
```
  6.127 ms (24827 allocations: 4.76 MiB)

**Simulation 2**
If we change to a NEP-solver involving the higher derivatives with `compute_Mder`, we
obtain a much bigger difference:
```julia
nep=nep_gallery("dep0_sparse",100);
n=size(nep,1);
(λ,v)=quasinewton(nep,v=ones(n),λ=2,tol=1e-11,displaylevel=1,maxit=200)
```
```julia
dnep1=deflate_eigpair(nep,λ,v,mode=:Generic);
@btime mslp(dnep1,λ=1im,tol=1e-11)
```
  44.163 ms (3188 allocations: 18.30 MiB)
```julia
dnep2=deflate_eigpair(nep,λ,v,mode=:SPMF);
@btime mslp(dnep2,λ=1im,tol=1e-11)
```
  43.594 ms (4020 allocations: 12.08 MiB)
Old approach:
[...]

PR #169, opened 2019-03-08
Jd update

eringh:
Quite some of the handling around deflation was previously done in the JD-code. But as of the deflation rewrite #167 that is no longer needed (the new code also seems to be more efficient). Hence that overhead is removed from JD.   (Also picks up an update from #111, when in changing in the JD-code)


eringh:
Should I move the Schur-complement LinSolver for deflated NEPs to `LinSolvers`? As discussed in #167.

jarlebring:
Wow. It really reduces the code a lot. Nice!

src/method_jd.jl
```diff
@@ -257,18 +250,13 @@ function jd_effenberger(::Type{T},
     end
 
     conveig += 1
-    Λ[1,1] = λ
-    X[:,1] = u
-    deflated_nep = DeflatedGenericNEP(nep, Λ[1:conveig,1:conveig], X[:,1:conveig])
+    deflated_nep = deflate_eigpair(nep, λ, u)
```
jarlebring: Would it be possible to pass the deflation `mode` parameter as an optional parameter to the JD-call?

jarlebring:
Yes. I think the Schur-complement linsolver can be moved. I changed my mind again: I think it belongs in `nep_deflation.jl` not LinSolvers. Maybe it should be the `default_linsolver` for deflated  `DeflatedNEP`?

eringh:
It turns out we have some code overlap also in the linear systems. The function `compute_U` has in part the same functionality as some of the code in `compute_Mder` for deflated NEPs. I think we should align and try to work on code-reuse when moving the linear solver.

jarlebring:
Yes. It sounds sensible to separate the U-part in the `compute_Mder(::DeflatedGenericNEP...)` (where it is referred to as `Q`) into a separate function and then call that function also in  `compute_U`.

eringh:
After new discussion: We do not implement a generic Schur complement solver for saddle point matrices. The feature could turn out a bit complicated and is currently only needed in JD, where it is already implemented.

eringh:
I feel ready to merge when the tests are done.
@jarlebring: What do you think?

jarlebring:
It looks good. Ready for merge.

Issue #154, opened 2019-01-08
Removing compute_eigvec_from_eigval 


jarlebring:
In NEPSolver.jl we currently have two functions `compute_eigvec_from_eigval_lu` and `compute_eigvec_from_eigval_lopcg`. They originate from a situation when the user specified a shift which was too close to an eigenvalue. 

These functions / functionality is incomplete in many ways, and I think the problem they solve is not so important (since a user can anyway select a different shift). I suggest we remove these and replace the calls in the NEPSolvers with generating an error. The best code is no code at all https://blog.codinghorror.com/the-best-code-is-no-code-at-all/

eringh:
After the removal from Beyn contour method it seems as if these functions are used in:
- `newton`
- `augnewton`
- `quasinewton`
- `newtonqr`
- `implicitdet`
- `resinv`
- `rfi`
- `rfi_b`

Is it always a shift too close to the eigenvalue that causes the singular case, can it not also be that the current iterate it too close to an eigenvalue but we have not yet computed a good enough approximation to the eigenvector? In that case, should we throw an error, or just a warning, or perhaps let the user specify?

jarlebring:
The problem that we aimed to solve was that if the shift was already an eigval, then we received an error. We preferred that the method returns an eigpair. In the interest of reducing the code-base I think this should be reconsidered. Letting the user provide what happens seems like [feature creeping](https://en.wikipedia.org/wiki/Feature_creep), and would generate more code.

The current solution captures some cases but not all.  Sometimes LAPACK throws an `LAPACKUnknwonError` if the matrix is singular and we only capture `SingularException`. So the current solution is hackish, does not cover all cases, and contains quite a bit of code.

I would say just let the error propagate all the way to the user.

PR #171, opened 2019-03-11
Remove functions compute_eigvec_from_eigval

eringh:
Removes code with reference to the discussion in #154, as the `compute_eigvec_from_eigval` was quite a bit of code, did not cover all cases, and was a bit hacky.


jarlebring:
Nice! 

Merging this will have quite a few conflicts once #168  is merged... so if you need to do more modifications to this PR, it might be good to first sync it with the branch errmeasure_rewrite.

eringh:
I don't think this PR will contain much more. I made it this way to make sure I did not accidentally break something, and to document if we want to go back. We can also wait with merging this until #168 is merged to master (since the latter is more complicated and hence probably nicer to not get merging conflicts). How long do you think that will take?

jarlebring:
:+1:

eringh:
Conflicts resolved. Unfortunately the diff-tool is not very good at seeing what has happened (so the conflicts were also quite messy to resolve). As soon as the tests are ok I'll merge to master.

eringh:
The merge was not nice and this seems to messy to clear out. I will create a new branch from master and do the changes again. It was not that difficult, mostly manual work.

PR #172, opened 2019-03-12
Removed compute_eigvec_from_eigval-functions.

eringh:
Removes code with reference to the discussion in #154, as the compute_eigvec_from_eigval was quite a bit of code, did not cover all cases, and was a bit hacky. Replaces #171 since that merge was too messy.


jarlebring:
:+1:

PR #176, opened 2019-04-21
New method: higher order moment contour integration (asakura sakurai)

jarlebring:
A first version of higher moment contour integration: `contour_assu`. 

``` 
julia> nep=SPMF_NEP([[0 1 ; 1 1.0], [1 0 ; 0 0]], [s->one(s),s->exp(1im*s^2)]);
julia> λ,V=countour_assu(ComplexF64,nep,radius=3,neigs=6)
julia> minimum(svdvals(compute_Mder(nep,λ[1])))
1.2824282172172952e-14
julia> minimum(svdvals(compute_Mder(nep,λ[4])))
2.271258726650263e-14
```

Reference: Asakura & Sakurai  https://doi.org/10.14495/jsiaml.1.52

Some things currently missing in this PR.

1. The specification of quadrature method is not as general as in `contour_beyn`. Ideally we want code sharing. At the moment I don't know the formulation of the quadrature functions, without adding additional linear solves. Might need rewriting of the quadrature functions in `contour_beyn`.

2. Eigenvector extraction: Need to read up some more 

3. Removal of bad eigenvalues (as in  `contour_beyn`). Should be straightforward once we have eigvec extraction.

4. Only disk supported. This is how the method is usually expressed. Need to go through the derivation again to get the coefficients in the integrals correct to be generalizable to e.g. ellipses

### DIFF
--- a/src/NEPSolver.jl
+++ b/src/NEPSolver.jl
@@ -40,6 +40,7 @@ Executes z if displaylevel>1.
     include("method_rfi.jl")
     include("method_jd.jl")
     include("method_beyncontour.jl")
+    include("method_assucontour.jl")
     include("method_blocknewton.jl")
     include("method_broyden.jl")
     include("method_ilan.jl")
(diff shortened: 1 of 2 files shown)


jarlebring:
Any thoughts on point 1 @wrs28 ?

Issue #178, opened 2019-06-13
TIAR invalid index: 1.0 of type Float64


jarlebring:
Tiar throws a strange error when `Neig=inf`
```
julia> nep=nep_gallery("dep0",100);
julia> v0=ones(size(nep,1));
julia> λ,v=iar(nep;v=v0,tol=1e-5,Neig=Inf); # No problems
julia> λ,v=tiar(nep;v=v0,tol=1e-5,Neig=Inf); 
ERROR: ArgumentError: invalid index: 1.0 of type Float64
Stacktrace:
 [1] to_index(::Float64) at ./indices.jl:270
 [2] to_index(::Array{Int64,1}, ::Float64) at ./indices.jl:247
 [3] to_indices at ./indices.jl:298 [inlined]
 [4] to_indices at ./indices.jl:295 [inlined]
 [5] getindex at ./abstractarray.jl:927 [inlined]
 [6] macro expansion at ./multidimensional.jl:671 [inlined]
 [7] macro expansion at ./cartesian.jl:64 [inlined]
 [8] macro expansion at ./multidimensional.jl:666 [inlined]
 [9] _unsafe_getindex! at ./multidimensional.jl:662 [inlined]
 [10] _unsafe_getindex at ./multidimensional.jl:656 [inlined]
 [11] _getindex at ./multidimensional.jl:642 [inlined]
 [12] getindex(::Array{Int64,1}, ::StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}) at ./abstractarray.jl:927
 [13] #tiar#24(::Type{IterativeSolvers.DGKS}, ::Int64, ::typeof(default_linsolvercreator), ::Float64, ::Float64, ::Type{DefaultErrmeasure}, ::Complex{Float64}, ::Complex{Float64}, ::Array{Float64,1}, ::Int64, ::Int64, ::Bool, ::Type, ::typeof(tiar), ::Type{Complex{Float64}}, ::DEP{Float64,Array{Float64,2}}) at /home/jarl/.julia/packages/NonlinearEigenproblems/1dPIC/src/method_tiar.jl:202
```
Seems to be related to the partial implementation of the feature in #177. (`Neig=100` does not throw an error.)

eringh:
I think the problem is the type.
```
julia> typeof(Inf)
Float64
julia> typeof(min(4,Inf))
Float64
```
Because I can recreate a similar error as:
```
[...]

Issue #177, opened 2019-06-13
Unknown nof eigvals-mode for arnoldi methods


jarlebring:
Currently, iar, tiar, nlarnoldi, etc take the number of wanted eigenvalues as input: 
```
iar(nep,Neig=3)
```
If `Neig` eigenvalues are not found, an error is thrown. When working on a specific application, I realize the fact that I don't know how many eigenvalues I want beforehand. Instead, I would like it to run  `maxit` iterations and just return as many eigenvalues as have converged at that point.

Feature request: We let this be the behavior if `Neig=Inf`. The only difference would (essentially) be that we do not throw an error after `maxit` iterations, but we just return the currently computed solutions. 

This seems to already (partially?) implemented if we look at the end of `iar`. It is however not activated in tiar. 

It should also be documented in `?iar` etc

eringh:
Should this be done also for _infinite Lanczos_ och _infinite bi-Lanczos_? In such case, should we consider doing some kind of code-reuse functionality? Right now there is almost the same code in a few places (the bug in #178 has been so far been corrected in three different methods).

eringh:
_Infinite Lanczos_ (`ilan`) could possibly also get this behavior. Is that desired, @meleg?

jarlebring:
Minor. I think you can just fix it, without disturbing the parental leave.

eringh:
Are there any other methods we should consider having this behavior in? I am not sure if `nlar` or `jd ` are suitable for it. It could possibly be done, but with a penalty in the pre-allocated memory.

PR #179, opened 2019-06-14
Inf number of target eigenvalues in IAR and TIAR

eringh:
Fixes a bug reported in #178 where `Neig=Inf` gives a type error (fixed in IAR, TIAR, and IAR_CHEB). Also works on a feature requested in #177, where  `Neig=Inf` should be interpreted as running `maxit` iterations and return all converged Ritz pairs.


eringh:
I suggest merging soon to close #178. Further updates discussed in #177 can be taken separately (e.g., ILAN and IBiLAN).

jarlebring:
Looks good. :+1:

PR #180, opened 2019-06-14
Infinite Bi-Lanczos fix

eringh:
Fixes some bugs in Infinite Bi-Lanczos:
- `maxit` number of iterations was not possible due to an off-by-one error
- `NoConvergenceException` was not thrown due to how variables were defined

Also adds feature compute as many eigenvalues as possible in `maxit` iterations. As discussed in #177.


jarlebring:
Nice fix. From my perspective, it can be merged.  :+1:

Issue #170, opened 2019-03-09
Logging and debug output


jarlebring:
We often want to present diagrams with convergence as a function of iteration or CPU-time. This is currently not so easy. This is a bit the same problem that all the debug printouts for the methods are different. Therefore, I think there is a need for a logging system. 

Edit: What is described below is not quite how it ended up. See #182 for newer description.

I propose the following setup:

**User** 
You instantiate a logger object before calling the algorithm and hand it over to the NEP-solver as a kwarg
```julia
mylogger=IterationLogger()
rfi(nep,logger=mylogger);
```
After the iteration we can extract error history or other iteration information:
```julia
(iteration,errhist)=logger.get_iteration_history()
```
The logger would decide both what is stored in the log and what is printed. The Loggers can have a kwarg displaylevel, which essentially takes the role of the previous displaylevel. So this would make printouts similar to before:
```
rfi(nep,logger=NullLogger(displaylevel=1));
```

**NEP developer** 

As a method developer, you would have the logger kwarg and no displaylevel kwarg
```julia
function rfi(nep::NEP,logger::NEPLogger=NullLogger())
```
We need several ways to put things into the log (and/or printout). We need to be able to add text entries into the log:
```
push!(logger,"Starting LU-solve");
```
which can be easily implemented in the solvers by changing the `@ifd` macro. We also need ways to add information about iteration history:
This would store the error of eigenvalue `s` at iteration `i`:
```
push_iteration_info!(logger,i,err,s,v);
```
The `push_iteration_info!` would potentially also make a printout (if displaylevel>0) with eigpair information. This would make the presentation of iteration history consistent for all methods.

PR #182, opened 2019-06-17
Implementation of Logger 

jarlebring:
This is an attempt to incorporate a logging functionality #170.  

As a user: You can use the new `logger` kwarg for the NEP-solver:
* If `logger` is an Int, it will behave as displaylevel before, under the hood a `PrintLogger` is instantiated.
* If `logger <: Logger` one can obtain more detailed control of what is printed and one can store iteration info in the `logger` object.

As a NEP-method developer:
* `push_info!(logger,"text")` can be used to dump info to the logger. If the logger is a PrintLogger "text" will just be printed. Note that you can add variables by using the dollar sign: "x=$x"  
* `push_iteration_info(logger,iter,kwargs...)` can be used to push information about iteration `iter`. kwargs include things such as `err`, `v`, etc

* [x] method_newton 
* [x] iar tiar 
* [x] ErrorLogger storing information
* [x] rfi
* [x] nlar
* [x] infbilanczos
* [x] contour integrals
* [x] nleigs 
* [x] sgiter
* [x] mslp
* [x] jd
* [x] ilan
* [x] broyden (w inner solves)
* [x] block newton
* [x] unit tests
* [x] comments + documentation

Done with: a level parameter:  ~~At the moment I am not sure how to handle `iar`~~ ~~and other methods which compute many eigenvalues. Just printing all the eigval approximations by default would be give too many printouts, but we want this information when we want to study the method (e.g convergence diagrams).~~

### DIFF
--- a/docs/src/bemtutorial.md
+++ b/docs/src/bemtutorial.md
@@ -83,7 +83,7 @@ nep=BEM_NEP(mesh,gauss_order);
 After creating the NEP, you can try to solve the problem with methods in the package, e.g.,
 [`MSLP`](methods.md#NonlinearEigenproblems.NEPSolver.mslp) works quite well for this problem:
 ```julia-repl
-julia> (λ,v)=mslp(nep,λ=8,displaylevel=1)
+julia> (λ,v)=mslp(nep,λ=8,logger=1)
 Iteration:1 errmeasure:4.122635537095636e-6 λ=8.128272919317748 + 0.007584851218214716im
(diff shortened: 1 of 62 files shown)


jarlebring:
Okay. Starting to feel ready with this PR.

eringh:
I agree. It starts to be ready to merge.

jarlebring:
I like it. Time to merge.

PR #189, opened 2019-07-19
Contour block SS

jarlebring:
New attempt for the contour integral method block SS method (previously contour_assu) described in [Asakura, Sakurai, Tadano, Ikegami, Kimura, A numerical method for nonlinear eigenvalue problems using contour integrals, JSIAM Letters, 2009 Volume 1 Pages 52-55](https://www.jstage.jst.go.jp/article/jsiaml/1/0/1_0_52/_article) 

Example:
```julia
julia> nep=SPMF_NEP([[0 1 ; 1 1.0], [1 0 ; 0 0]], [s->one(s),s->exp(1im*s^2)]);
julia> λ,V=contour_block_SS(nep);
julia> norm(compute_Mlincomb(nep,λ[1],V[:,1]))
1.3148059621411757e-15
```

There were some details missing in the original JSIAM-paper, so I had to use techniques from other papers. 

Still not quite finished: Cleanup needed.  We could also add other quadrature rules and ideally do code sharing with beyn,  but it requires quite a bit of reordering of the code. 

* [x] Change beyn quadrature to use Juliaesque types (and make it extendible)
* [x] quadrature code sharing with beyn
* [x] feature: integration over ellipse in block_SS
* [x] cleanup: function comments 
* [ ] ~~cleanup: make warnings like in beyn~~
* [x] tutorial on user-specified quadrature rule


jarlebring:
This new way of doing quadrature might break use of `beyncontour` for some people (e.g. work of @wrs28?)  but it allows a user to write their own integrator and it has a lot of code sharing btw block_SS and beyn.

jarlebring:
The unification of handling of error messages requires some more thought process and deserves a different PR.

jarlebring:
I'm now feeling more or less ready with this PR. The CI-failure seems to be due to a change of computers. I think travis now runs on a computer with less memory and fails for the WEP-tests. Unrelated to this PR.

PR #210, opened 2019-09-06
WEP update

eringh:
Updates in the Waveguide eigenvalue problem (WEP) in Gallery:
* Some cleanup of unused code
* A more Juliaesque implementation of some parts: function passing -> dispatch
* Overloading of some basic default linsolvers to make it easier to run a first test
* Updating of the large reference test


jarlebring:
Nice cleanup!

src/gallery_extra/waveguide/Waveguide.jl
```diff
@@ -489,11 +489,15 @@ The `kwargs` keyword argument is passed to the solver.
...
-        function WEPLinSolverCreator(;solver_type=:backslash,kwargs=())
+        function WEPLinSolverCreator(;solver_type=:factorized,kwargs=())
             return new(solver_type,kwargs);
         end
     end
 
+    # Overload Backslash and Factorize lindsolvers to make small WEP examples work out-of-the-box
+    create_linsolver(creator::BackslashLinSolverCreator,nep::WEP,λ) = create_linsolver(WEPLinSolverCreator(solver_type=:backslash),nep, λ)
+    create_linsolver(creator::FactorizeLinSolverCreator,nep::WEP,λ) = create_linsolver(WEPLinSolverCreator(solver_type=:factorized),nep, λ)
+
```
jarlebring: This type of construction cause difficult-to-find bugs in julia. To change the behaviour of `BackslashLinSolverCreator` such that it does not only do backslash but something else is a bit unnatural / dirty. If the reason for these lines is only to make DefaultLinSolverCreator work, I would suggest we remove them (am I right?)  I don't think we need to take beginners  into account for this specialized NEP.

eringh: I think you are right, I'll remove it. I was just unhappy with the resulting error (which tells the user to implement `compute_Mder`).

jarlebring: Indeed. Those error messages are a bit off. Can't you just implement a `compute_Mder(::WEP,..)` which throws an error?

eringh: Great idea! I'll fix it.

eringh: I did the error message so that it points the user towards `WEPLinSolverCreator`. So now there is bot the tests to look at, and a more comprehensive error message.

PR #211, opened 2019-09-06
Cork framework

meleg:
The pull request contains:

- New structure `CORK_pencil` to represent the pencil used in CORK
- Constructors of the of the nleigs and truncated iar pencils in the `CORK_pencil` format
- A function that build the pencil from a `CORK_pencil`

In order to do this a clean way, I needed to separate `nep_transformations.jl` from  `NEPTypes.jl `. Now the module `NEPTransformations.jl ` contains the functions in `nep_transformations.jl`.

An example of the usage of these new features is in the script `CORK_working_dir/test_comparison.jl`or just run this command 
`include("CORK_working_dir/test_comparison.jl")`

This pull request is very premature, but I can use it to receive suggestions and feedback.


jarlebring:
The file `CORK_working_dir/test_comparison.jl` seems to be missing. I will check the other in the meantime.

meleg:
This is the content of that file

```
using NonlinearEigenproblems, Random, LinearAlgebra, PyPlot, Revise

nep=nep_gallery("dep0");
unit_square = float([1+1im, 1-1im, -1-1im,-1+1im])

r=5; θ=range(0,stop=2π,length=1000); Σ=r*cos.(θ) + 1im*r*sin.(θ); Ξ=[-10.0]
# nleigs linearization
cp=compute_CORK_pencil(nep,NleigsCorkLinearization(Σ=Σ,Ξ=[-10.0]))
AA,BB=build_CORK_pencil(cp)
λ2=eigvals(Matrix(AA),Matrix(BB))

# iar linearization
cp=compute_CORK_pencil(nep,IarCorkLinearization(d=40))
AA,BB=build_CORK_pencil(cp)
λ3=eigvals(Matrix(AA),Matrix(BB))

# reference eigenvalues
λ,_=iar(nep,maxit=200,tol=1e-8,neigs=Inf)

pygui(true)
plot(real(λ),imag(λ),marker="*",markerfacecolor=:none,c=:red,linestyle=:none)
plot(real(λ2),imag(λ2),marker="o",markerfacecolor=:none,c=:red,linestyle=:none)
plot(real(λ3),imag(λ3),marker="d",markerfacecolor=:none,c=:red,linestyle=:none)
```
I will try to upload it later.

jarlebring:
It looks nice!

Some comments:
* Maybe you can consider giving the `XXLinearization`-types a different name, or use some other structure/functions. Just from the name, I would expect that a linearization-object also contain the data. In your case it only contains parameters for the linearization, not problem data. In a first implementation, I would actually just to remove the Linearization-types, and put the parameters as kwargs in the `compute_CORK_pencil()` (and separate the two `compute_CORK_pencil` to different function names). 
* As a verification, I would propose you plot the error of (e.g. one eigenvalue) as a function of terms in the expansion, e.g. `d` in IAR. You want to observe superlinear convergence. 
* How much have you experimented with low-rank CORK-linearization? I would try this quite early, since it has a lot to do with the `CORK_pencil`-type which is important (and more important than getting the nleigs-connection correct from the start in my opinion). Note that low-rank IAR also fits into low rank CORK. Low-rank iar:  http://dx.doi.org/10.1002/nla.2043

src/NEPTransformations.jl
```diff
@@ -0,0 +1,294 @@
...
+        return PEP(A)
+    end
+
+    # representation of the structured linearizations used in the CORK framework
+    struct CORK_pencil{T<:AbstractMatrix}
+        M::T
+        N::T
+        Av::Vector{T}   # Array of Array of matrices
+        Bv::Vector{T}   # Array of Array of matrices
+    end
```
jarlebring: I think you want the matrix-types to be decoupled. If the problem is a full matrix, you may still want to represent the other matrices in sparse. 
```
struct CORK_pencil{TM<:AbstractMatrix,TN<:AbstractMatrix,TAv<:AbstractMatrix,TBv<:AbstractMatrix}
    M::TM
...
```

jarlebring:
I don't really see why you changed the file `nep_transformation.jl` to `NEPTransformations.jl` and changed it to a module. It seems unrelated to CORK and this PR. Since you have done it, okay, I wouldn't reverse it. Please check that `iar_chebyshev` works as it should. Some functionality depends on `shift_and_scale`, which is now in a different module.

jarlebring:
I suggest to rename: `CORK_pencil` -> `CORKPencil`. We prefer camel case for types unless it get unreadable.

src/NEPTransformations.jl
meleg: We have discussed this issue a couple of times IRL and you convinced me, but I have now changed my mind during the implementation. The reason is the following:

Each matrix in the vector `Av` and `Bv` are computed as linear combination of matrices generated with `compute_Mder`. Therefore it makes sense that the type of `Av` and `Bv` is inherited by the type of the `nep`. So, if the `nep` is sparse (or not), so it will be the matrices in `Av` and `Bv`.

Same reasoning for the matrices `M` and `N`. They should have the same type of the `nep`. Indeed, if the linearization is explicitly computed, via `build_CORKPencil`, then if the `nep` is sparse, we want the linearization to be sparse. The linearization involve tensor products with the matrices `M` and `N`. Therefore they also should inherit the type from the `nep`.

Do you agree or you have a better argument to decoupled the matrix-types?

meleg:
> I don't really see why you changed the file `nep_transformation.jl` to `NEPTransformations.jl` and changed it to a module. It seems unrelated to CORK and this PR. Since you have done it, okay, I wouldn't reverse it. Please check that `iar_chebyshev` works as it should. Some functionality depends on `shift_and_scale`, which is now in a different module.

 `iar_chebyshev` (and the other methods) works as it should.

meleg:
I would like to merge to master. We can open an Issue with a check list of things to do. It is getting problematic to stay on this branch since `NEPTransformations` is now a module in this branch and I get continuously conflicts when I merge the master where `NEPTransformations` is not a module.

src/NEPTransformations.jl
jarlebring: I agree that we can have `eltype(Av)=eltype(Bv)`. 

I don't agree with this:
> Same reasoning for the matrices M and N. They should have the same type of the nep. Indeed, if the linearization is explicitly computed, via build_CORKPencil, then if the nep is sparse, we want the linearization to be sparse. The linearization involve tensor products with the matrices M and N. Therefore they also should inherit the type from the nep.

It might be okay to enforce `typeof(M)=typeof(N)` but not `typeof(M)=eltype(Av)`.   

Take the IAR-equivalent version. Then M and N will be extremely sparse matrices (independent if `Av` are sparse or not). If you couple `Av` and `N` you put limitations on the data type. 

That "the linearization is sparse" is not a good reasoning for a data type. The `CORKPencil` represents a  *factorization* of a linearization. The data type representing the factorization should be memory efficient.  If a *method* wants to construct the matrices corresponding to the `CORKPencil,` in a way that requires the types to be the same, the *method* should only be implemented for equal types, or (better) convert sparse to full matrices. Not all methods need to construct the actual pencil matrices.

jarlebring:
Merge is okay with me (although there is quite a bit of clean-up to do later).

Issue #188, opened 2019-07-09
Test fail on 1.2.0-rc


jarlebring:
Tests failing on the julia 1.2.0 release candidate:

* [ ] test deflation modes (new)
* [ ] test interpolation

```
Running test deflation (9 / 49)
[ Info: Computing eigenvalue k=1
[ Info: Computing eigenvalue k=2
[ Info: Computing eigenvalue k=3
Deflation modes (new): Error During Test at /home/jarl/jobb_synced/src/neppack_master2/NonlinearEigenproblems.jl/test/deflation.jl:44
  Got exception outside of a @test
  No convergence: 'Number of iterations exceeded. maxit=30.' eigenvalue approx:-0.06646277598910488 + 0.0im, errmeasure:0.5124720777960479
  Stacktrace:
   [1] #augnewton#12(::Type{DefaultErrmeasure}, ::Float64, ::Int64, ::Float64, ::Array{Float64,1}, ::Array{Float64,1}, ::Int64, ::typeof(backslash_linsolvercreator), ::Float64, ::Int64, ::typeof(augnewton), ::Type{Complex{Float64}}, ::DeflatedGenericNEP) at /home/jarl/jobb_synced/src/neppack_master2/NonlinearEigenproblems.jl/src/method_newton.jl:349
   [2] #augnewton at ./none:0 [inlined]
   [3] #augnewton#11 at /home/jarl/jobb_synced/src/neppack_master2/NonlinearEigenproblems.jl/src/method_newton.jl:273 [inlined]
   [4] (::getfield(NonlinearEigenproblems.NEPSolver, Symbol("#kw##augnewton")))(::NamedTuple{(:v, :λ, :tol),Tuple{Array{Float64,1},Float64,Float64}}, ::typeof(augnewton), ::DeflatedGenericNEP) at ./none:0
   [5] top-level scope at /home/jarl/jobb_synced/src/neppack_master2/NonlinearEigenproblems.jl/test/deflation.jl:73
   [6] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Test/src/Test.jl:1113
   [7] top-level scope at /home/jarl/jobb_synced/src/neppack_master2/NonlinearEigenproblems.jl/test/deflation.jl:46
   [8] include at ./boot.jl:328 [inlined]
   [9] include_relative(::Module, ::String) at ./loading.jl:1094
[...]

PR #218, opened 2019-09-10
Gallery matrices

eringh:
Attempt to address #188. So far it is only the random generator that are implemented. Gallery is not yet changed, and should be before we merge.

### DIFF
--- a/src/gallery_extra/GalleryRand.jl
+++ b/src/gallery_extra/GalleryRand.jl
@@ -35,14 +35,14 @@ export reset_rng!
         x::UInt128
         w::UInt128
         s::UInt128
-        function MSWS_RNG(seed::UInt128 = 0x00000000000000000000000000000000)
+        function MSWS_RNG(seed::UInt128 = zero(UInt128))
             this = new();
             reset_rng!(this, seed);
             return this
         end
     end
 
-    reset_rng!() = reset_rng!(GLOBAL_MSWS_RNG, 0x00000000000000000000000000000000);
+    reset_rng!() = reset_rng!(GLOBAL_MSWS_RNG, zero(UInt128));
     reset_rng!(rng::MSWS_RNG, seed::Integer) = reset_rng!(rng, unsigned(Int128(seed)));
     reset_rng!(seed::Integer) = reset_rng!(GLOBAL_MSWS_RNG, seed);
     reset_rng!(seed::UInt128) = reset_rng!(GLOBAL_MSWS_RNG, seed);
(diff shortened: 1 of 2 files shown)


jarlebring:
Nice work! You have a point that you maybe went overboard with types. I think we should aim for a much more minimalistic approach. You have implemented a number of user-friendly features, which generates quite a bit of code, with potential bugs and the code is already a bit complicated. Note that we (NEP-PACK developers) will be the only users of this, and these functions should not be visible to NEP-PACK users, and it is therefore okay if it is not extremely user friendly. In order to reduce the code, I suggest we remove the feature of global random seed, and only support Int64 and Float64 (complex can be generated by two calls by the gallery developer). I don't understand why the upper and lower parameters are needed in so many places. If we have a routine for [0,1] generation, can't you scale it to any interval? Integers for a given interval can be generated by rounding, e.g., `round(10+3*rand())`.  I think it is perfectly fine to push the rescaling to the "gallery problem developer". In the interest of reducing code: I suggest we remove functions whose only purpose is to throw error messages like "this should not happen". 

I also think we don't need unit tests for this (but now during development maybe it makes sense). Unit tests are mostly to verify features that the package supports. A NEP-PACK user is not expected to be able to call these functions.

eringh:
I can try to take away some of the code. However, regarding the global seed I think it is important since it is a way to make sure that calls of the type
```
A1 = rand(n,n)
A2 = rand(n,n)
```
do not generate the same matrices twice. The alternative would be to set up a `MSWS_RNG` struct pass it around in the gallery functions:
```
rng = MSWS_RNG()
A1 = rand(rng,n,n)
A2 = rand(rng,n,n)
```
The design is inspired by what they do in[ Random.jl](https://github.com/JuliaLang/julia/blob/c6da87ff4bc7a855e217856757ad3413cf6d1f79/stdlib/Random/src/RNGs.jl#L295)

The integers on an interval can probably be generated with a rounding approach. I went for an approach were the operations stays on integers, but that might not be necessary.

jarlebring:
Regarding seed: I think your second option is fine. We can demand that the
gallery developer takes care of the seed.

eringh:
To get the code complexity down, the code currently only supports matrices with `Float64`. However, I realized that it will be difficult to generate a sparse complex matrix, since there is no longer a complex draw and `A+1im*B` might not be so nice for sparse matrices.

jarlebring:
I think we don't need random sparse gallery problems. And if we do, we can
leave it to the gallery problem developer to solve.

jarlebring:
I meant we don't need *complex* sparse problems.

eringh:
Regarding global seed: I can remove it, but it will only remove like 9 lines of code. Are there any other reasons to remove it?

jarlebring:
Nice. I made a checklist on the top since there are so many we may not want to do them all in one go.

Issue #225, opened 2019-09-23
computers with different wordsize  (Raspberry Pi problem)


jarlebring:
I tried to run our package on raspberry pi. It works mostly. 

However, we have sometimes incorrectly assumed that we always have `Int64=Int`. This is not true for 32-bit systems:
```julia
jarl@pirog:~ $ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.0.3
 _/ |\__'_|_|_|\__'_|  |  Raspbian ⛬  julia/1.0.3+dfsg-4+rpi1
|__/                   |

julia> Int  # This returns Int64 on most systems we use
Int32

julia> versioninfo()
Julia Version 1.0.3
Platform Info:
  OS: Linux (arm-linux-gnueabihf)
  CPU: ARMv7 Processor rev 4 (v7l)
  WORD_SIZE: 32
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, cortex-a53)
Environment:
  JULIA_LOAD_PATH = .:/home/jarl/archive_synced/scripts:

shell> lsb_release -a
No LSB modules are available.
Distributor ID: Raspbian
Description:    Raspbian GNU/Linux 10 (buster)
Release:        10
Codename:       buster
```
(It does not seem to matter for floats.)

It's mostly okay since we mostly use `Int` consistently, but not always, so it  does cause some problems in those rare occations we use `Int64`.
```julia
julia> nep=nep_gallery("dep0",100);
ERROR: MethodError: no method matching gen_rng_mat(::NonlinearEigenproblems.Gallery.MSWS_RNG, ::Int32, ::Int32)
Closest candidates are:
  gen_rng_mat(::NonlinearEigenproblems.Gallery.MSWS_RNG, ::Int64, ::Int64) at /home/jarl/.julia/packages/NonlinearEigenproblems/NAm4H/src/gallery_extra/basic_random_examples.jl:98
Stacktrace:
 [1] dep0(::Int32) at /home/jarl/.julia/packages/NonlinearEigenproblems/NAm4H/src/gallery_extra/basic_random_examples.jl:4
[...]

PR #226, opened 2019-09-24
Integer updates

eringh:
Improves running on 32-bit system by updating ussage of `Int64`. See #225


jarlebring:
Since we have stored serialized files in the repository in `Int64`, I think the default for both reading and writing should be `Int64` not `Int`.

jarlebring:
:+1:

PR #228, opened 2019-10-03
Ilan testing

meleg:
In this PR I have:
1. Improved `ilan`, e.g., now it is simpler to specify  the `inner_solver_method`
2. Tested more `ilan`

Here you find three examples of ilan for solving a DEP, the gun problem and the symmetrization of a random problem.

[ILAN_scripts.zip](https://github.com/nep-pack/NonlinearEigenproblems.jl/files/3685801/ILAN_scripts.zip)


src/method_ilan.jl
```diff
@@ -112,16 +114,16 @@ function ilan(
     a=Vector{T}(γ.^(0:2m+2)); a[1]=zero(T);
     local M0inv::LinSolver = create_linsolver(linsolvercreator,nep,σ)
     err=ones(m,m);
-    λ=zeros(T,m+1);
+    λ=zeros(T,5m+1);
```
jarlebring: Why 5m? Is 5 arbitrary? Maybe it should be a solver parameter?

src/method_ilan.jl
```diff
@@ -166,32 +172,71 @@ function ilan(
...
-            # solve the projected NEP
-            push_info!(logger,2,"Solving the projected problem",continues=true);
-            λ,ZZ=iar(pnep;neigs=Inf,logger=0,maxit=150,tol=tol,check_error_every=Inf,errmeasure=err_lifted)
-            push_info!(logger,2,".");
-            W=VV*ZZ;
+                q=length(λproj)
+                if q>m q=m end
+                mul!(view(W,:,1:q),VV,view(Wproj,:,1:q))
+                #ZZ=VV*Zproj[:,1:k];
+                λ=λproj[1:q];
```
jarlebring: Doesn't this line overwrite the allocated λ?  Maybe the  `λ=zeros(T,5m+1);` above can be removed

src/method_ilan.jl
meleg: It should be 2m+2, I will fix it.

src/method_ilan.jl
```diff
@@ -76,6 +76,8 @@ function ilan(
     check_error_every=30,
     inner_solver_method=DefaultInnerSolver(),
     Compute_Bmul_method::Type{T_y0}=Compute_Bmul_method_Auto,
```
jarlebring: I think the kwarg `Compute_Bmul_method` should start with a small letter. I know it can be discussed if it is a type and therefore a capitalized. It is a kwarg parameter and for a user I think it is more natural to specify a kwarg with a small letter. This is also how it is done in other parts of the code.

src/method_ilan.jl
```diff
@@ -131,9 +133,12 @@ function ilan(
     Av=get_Av(nep)
```
jarlebring: In the documentation of the solver, remember to say that the solver must be an SPMF (or at least have an implementation of `get_Av`=.

src/method_ilan.jl
```diff
@@ -201,13 +246,22 @@ function ilan(
     end
```
jarlebring: The lines
```
        Qp=Q;   Q=Qn;
        Qn=zero(Qn);
```
allocate quite a bit of memory. Don't you want to do inplace copy?  Smth like
```
 Qp[:]=Q;
 Q[:]=Qn; 
 Q[:]=0;
```

jarlebring:
The printout is incorrect when `proj_solve=true`. Compare:
```julia
julia> nep=nep_gallery("dep_symm_double",10);
julia> v0=ones(size(nep,1));
julia> λ,v=ilan(nep;v=v0,neigs=3,logger=1,check_error_every=1,maxit=50,proj_solve=false);
---
----
-----
------
-------
--------
---------
----------
+----------
+-----------
+------------
+-------------
+--------------
+---------------
+----------------
+-----------------
+---=--------------
--+=----------------
-+---+---------------
--+--+----------------
--+--+-----------------
=-=--+------------------
++---+-------------------
julia> λ,v=ilan(nep;v=v0,neigs=3,logger=1,check_error_every=1,maxit=50,proj_solve=true);
-
-
-
-
-
-
-
-
-
+
+
-
+
-
=
-
+
-
-
-
+
-
-
+
-
-
+
-
-
+
+
=
+
+
-
-
+
+
-
+
-
=
-
+
-
-
=
-
```

Maybe it is a deeper problem...

src/method_ilan.jl
```diff
@@ -76,6 +76,8 @@ function ilan(
     check_error_every=30,
     inner_solver_method=DefaultInnerSolver(),
     Compute_Bmul_method::Type{T_y0}=Compute_Bmul_method_Auto,
+    proj_solve=true,
+    inner_logger=0
     )where{T<:Number,T_orth<:IterativeSolvers.OrthogonalizationMethod,T_y0<:Compute_Bmul_method}
 
     @parse_logger_param!(logger)
```
jarlebring: `@parse_logger_param!` should be called also for the inner logger, cf iar and other methods with inner solvers.

src/method_ilan.jl
```diff
@@ -166,32 +172,71 @@ function ilan(
 
         orthogonalize_and_normalize!(view(V,:,1:k),view(V,:,k+1), view(HH,1:k,k), orthmethod)
 
+
         # extract eigenpair approximation
         if ((rem(k,check_error_every)==0)||(k==m))&&(k>2)
```
jarlebring: Can the `k>2`-condition be removed? It has been removed from iar etc. For a user who wants to plot the error, this basically prevents you from accessing the error in iteration 1 and 2.

src/method_ilan.jl
```diff
@@ -112,16 +114,16 @@ function ilan(
     a=Vector{T}(γ.^(0:2m+2)); a[1]=zero(T);
     local M0inv::LinSolver = create_linsolver(linsolvercreator,nep,σ)
     err=ones(m,m);
```
jarlebring: This does not seem like a good way to allocate memory for error. Type is not set (should probably be `real(T)`). It is better to initiate it to NaN's.

jarlebring: The problem with output incorrect can be ignored (sorry). It seems to just be that the inner solver only returns one solution. This call works: `λ,v=ilan(nep;v=v0,tol=1e-5,neigs=3,maxit=50,logger=1,check_error_every=1,inner_solver_method=IARInnerSolver(maxit=50));`

meleg:
@jarlebring  Thanks for the careful reading and testing!
I am fixing the code and I think I have found a bug in `inner_solver`. 

I put oustide of the context the issue: we construct a  DEP and we want to check that it is a DEP
```
using NonlinearEigenproblems
nep=nep_gallery("dep0")
typeof(nep)==NEPTypes.DEP
```
This return false.
But if we check in another way it works
`isa(nep,NEPTypes.DEP)`
this return true.

I do not know why we have this. It should work both ways.

Anyway, according to what I just showed, I think we should change this function in `inner_solver`
```
function inner_solve(is::DefaultInnerSolver,T_arit::Type,nep::NEPTypes.Proj_NEP;kwargs...)
    if (typeof(nep.orgnep)==NEPTypes.PEP)
        println("case 1")
        return inner_solve(PolyeigInnerSolver(),T_arit,nep;kwargs...);
    elseif (typeof(nep.orgnep)==NEPTypes.DEP)
        println("case 2")
        # Should be Cheb IAR
        return inner_solve(IARChebInnerSolver(),T_arit,nep;kwargs...);
    elseif (typeof(nep.orgnep)==NEPTypes.SPMF_NEP) # Default to IAR for SPMF
        println("case 3")
        return inner_solve(IARInnerSolver(),T_arit,nep;kwargs...);
    else
        println("case 4")
        return inner_solve(NewtonInnerSolver(),T_arit,nep;kwargs...);
    end
end
```

jarlebring:
Indeed. Doing comparisons by equality signs is not good with types. If you check equality, the parameters in the parametric type have to be the same:
```
julia> struct Y{Z<:Number} a::Z; end
julia> Y(3.0)
Y{Float64}(3.0)
julia> typeof(Y(3.0))==Y
false
julia> typeof(Y(3.0))=Y{Float64}
true
```
`inner_solver(::DefaultInnerSolver,..)` should be changed to use `isa` instead.

PR #229, opened 2019-10-05
fixed inner_solve

meleg:
Edited  `inner_solver` according to the discussions in #228

### DIFF
--- a/test/inner_solves.jl
+++ b/test/inner_solves.jl
@@ -23,20 +23,21 @@ Random.seed!(0);
 
     logger=ErrorLogger(1,100,0);
     λv,V = inner_solve(DefaultInnerSolver(), ComplexF64, pnep; λv=[0.0,1.0] .+ 0im, neigs=3, V=ones(k, 2), tol=eps()*100, inner_logger=logger)
-    verify_lambdas(2, pnep, λv, V, eps()*500)
+    verify_lambdas(3, pnep, λv, V,eps()*10000)
     E=logger.errs[:,1]; E=E[(!isnan).(E)];
-    @test length(E) > 5 #Has done more than 5 iterations
-    @test E[end] < eps()*500 # Error measure fulfills stopping criteria
+    # THE FOLLOWING TWO TESTS ARE DISABLED: TO FIX
+    #@test length(E) > 5 #Has done more than 5 iterations
+    #@test E[end] <eps()*10000 # Error measure fulfills stopping criteria
 
     logger=ErrorLogger(1,100,0);
     λv,V = inner_solve(NewtonInnerSolver(), ComplexF64, pnep; λv=[0.0,1.0] .+ 0im, V=ones(k, 2), tol=eps()*100, inner_logger=logger)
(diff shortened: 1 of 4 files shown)


test/inner_solves.jl
```diff
@@ -23,20 +23,21 @@ Random.seed!(0);
...
     logger=ErrorLogger(1,100,0);
     λv,V = inner_solve(DefaultInnerSolver(), ComplexF64, pnep; λv=[0.0,1.0] .+ 0im, neigs=3, V=ones(k, 2), tol=eps()*100, inner_logger=logger)
-    verify_lambdas(2, pnep, λv, V, eps()*500)
+    verify_lambdas(3, pnep, λv, V,eps()*10000)
     E=logger.errs[:,1]; E=E[(!isnan).(E)];
-    @test length(E) > 5 #Has done more than 5 iterations
-    @test E[end] < eps()*500 # Error measure fulfills stopping criteria
+    # THE FOLLOWING TWO TESTS ARE DISABLED: TO FIX
+    #@test length(E) > 5 #Has done more than 5 iterations
+    #@test E[end] <eps()*10000 # Error measure fulfills stopping criteria
```
meleg: @jarlebring these tests did not pass and I am not sure I know how to fix them.

jarlebring: Actually, I think @eringh wrote those tests: abf771c1268cf9bda010a27ae5d5d791d682d22b. Since the Default for PEP should be a companion matrix, there are no iterations so these should probably be removed.

jarlebring:
All fine. For me the test for ErrorLogger + DefaultInnerSolver  can be removed / replaced.

test/inner_solves.jl
eringh: The nep we test it on is a projected `DEP`. What happened before was the Newton was used as inner solver (which just proves the point of the bug). Now instead IAR_Cheb is called (that is also visible in the fact that we now find 3 eigenvalues).

To reflect that the call in line 25 should maybe be changed to
`λv,V = inner_solve(DefaultInnerSolver(), ComplexF64, pnep; σ=0.0+0im, neigs=3, inner_logger=logger)`
since `λv` and `V` are just discraded by the inner solver anyways.

The problem with the test seems to be that IAR_Cheb does not push iteration info to the ErrorLogger in a good way. Is it supposed to be like that?

jarlebring: Thanks! This seems to be due to a issue, not `iar_chebyshev`. If you change the logger instantiation to `logger=ErrorLogger(50,100,0);` you get the error values.  In principle you can fix the test in that way. 

50 corresponds to the size of the buffer in terms of number of errors per iteration to store. If this is too small it seems to completely ignore the errors, which I think is wrong.

jarlebring: On second thought. This is okay, since there is a message saying that the buffer is full. This message is not visible in this case since it sets displaylevel=0.

meleg:
Is this fixed? Can we merge?

jarlebring:
Yes. It's okay with me.

Issue #227, opened 2019-09-26
Cork framework


meleg:
See checklist in #230

meleg:
This script is a verification for the IAR linearization. 

```
using NonlinearEigenproblems, Random, LinearAlgebra, PyPlot, Revise

nep=nep_gallery("dep0",50); n=size(nep,1)

err_vec=zeros(10)
for d=1:10
    cp=compute_CORKPencil(nep,IarCorkLinearization(d=d))
    AA,BB=build_CORKPencil(cp)
    λ,v=eigen(Matrix(AA),Matrix(BB))
    idx=argmin(abs.(λ)); λ1=λ[idx]; v1=v[1:n,idx]
    err_vec[d]=compute_resnorm(nep,λ1,v1)
    println(err_vec[d])
end
logC,p=[ones(9) log.(err_vec[1:end-1])]\log.(err_vec[2:end])
C=exp(logC)

println("The convergence order is: ",p)

pygui(true)
semilogy(err_vec,label="error")
# compute the fitted error
err_fit=zeros(10); err_fit[1]=err_vec[1]; for j=1:9 err_fit[j+1]=C*err_fit[j]^p end
semilogy(err_fit,linestyle="--",color="red",label="fitted error")
xlabel("degree PEP")
ylabel("Error")
legend()
```
The convergence follows `err_vec[j+1]=C*err_vec[j]^p`. The convergence order is `p=1.0221410654209089` and the constant is `C=0.0510161259959836`.
This is the output figure 
![figure_1](https://user-images.githubusercontent.com/33252231/65669684-516ce600-e044-11e9-9ab9-1d488c5cc0f5.png)

jarlebring:
Could you do this modifications in a PR, please.

PR #230, opened 2019-10-11
Cork lr

meleg:
In this PR the low-rank data structure for the CORK pencil is implemented. 

Here is the checklist: (transferred from #227)
- [x] matrix-types has to be decoupled in `CORK_pencil`
- [x]  better names to `XXLinearization`-types (or restructure it)
- [x] verification (plot the eigenvalue error as function of number of terms in the expansion)
- [x]  implement low-rank `COR_pencil_structure`
- [x]  documentation
- [x] tests
- [ ] remove / lessen duplication of code: `nleigs_coefficients.jl` and `method_nleigs.jl`  
- [ ] remove the file `nep_transformations.jl`  (check that all was transferred to NEPTransformations.jl)


jarlebring:
All checkboxes checked except two which can be done in a separate PR.

Issue #241, opened 2020-04-15
NEPSolver.nlar() not receiving λ0 parameter


carlomontec:
Seems NEPSolver.nlar() is not receiving the parameter λ0.  From the source code seems it should be called  λ?

jarlebring:
Thanks. It should follow the same naming as other solvers as we describe in the 
[solver developer style guidelines](https://github.com/nep-pack/NonlinearEigenproblems.jl/wiki/Style-guidelines-and-notes).

eringh:
Solved by #243.

PR #243, opened 2020-09-16
Style fix

eringh:
Update the style and fixes the bug in documentation, as reported in #241.


eringh:
I also saw that the `compute_rf_wrapper`s also take `λ0` as initial guess. Does not seem as serious since the wrappers are for more internal use than the interface of algorithms. But should I fix it while having this open?

jarlebring:
:+1:  for making `compute_rf` also consistent

eringh:
I think it is ready to merge. A search of the repo seems to reveal that no `λ0` are used in function definitions (although as temporary variables in tutorials and doc).

Issue #245, opened 2020-09-18
Tests break in version 1.5.1


eringh:
Some tests break in Julia version 1.5.1. Specifically:

- JD: Test fails (Singular exception)
- Inner_Solvers_ Tests fails
- NLEIGS particle r2: Test fails
- wep_small: Segfault

The three first i suspect are due to problems with some random initializations. Similar problems as in #188.
The segafault needs to be investigated separately.

eringh:
```
julia> include("wep_small.jl")
[ Info: Precompiling GalleryWaveguide [top-level]
[ Info: WEP default linsolver
Precomputing linsolver
iter 1 err:0.4389897220963379 λ=-3.0 - 3.5im
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception:
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION  with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception:
Please submit a bug repor
Please submit a b
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x1a5bb500 --  g report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x1a5bb500 -- unknown function (ip: 0000 at 0x1a5bb500 -- unknown function (unknown function (ip: 0000 at 0x1a5bb500 -- unknown function (ip: 000000001A5BB500)
  reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x1a5bb500 -- unknown function (ip: 000000001A5BB500)
in expression starting at C:\Users\emilrin expression starting at C:\Users\emilr\Documents\GitHub\NonlinearEigenproblems.jl\test\wep_small.jl:12
in expression starting at C:\Users\emilrin expression starting at C:\Users\emilr\Documents\GitHub\NonlinearEigenproblems.jl\test\wep_small.jl:12
in expression starting at C:\Users\emilr\Documents\GitHub\NonlinearEigenproblems.jl\test\wep_small.jl:12
 ir entirety). Thanks.
[...]

PR #246, opened 2020-09-18
Test fix for version 1.5.1

eringh:
The fixes for `Inner_solver` and `NLEIGS particle r2` are pretty uggly to say the least. However, it was the best I could come up with when "simple" deterministic vectors did not give passing tests.
One could start playing with new seed, but that felt pretty irritating since there is no guarantee that it will not change in future releases anyway. At least these initializations are stable.

I have to look further into the segfault of `wep_small ` before I am done.


eringh:
Given the discussion in #245, this PR seems to resolve the problems with NEP-PACK.
Some of the fixes are ugly. Do we need a stable function for random starting vectors? (Sometimes a vector of ones does not seem to work well.)
Or is it ready to merge?

PR #249, opened 2021-06-15
Fixed typo in documentation of REP (A0 -> A1)

harrymd:


jarlebring:
Thanks. The rational representation is not so mature and still needs some work, e.g. variations such with barycentric form etc.

